PR #55, opened 2014-12-18
Custom namespace for addons

nebulousGirl:
To make it possible to manually register addons from a Laravel package, I added a fourth parameter to the register method of the AddonBinder.

By doing so, one can now create an Addon outside of the "core" folder.

This is my personal take on Addons. I don't know how one would develop addons for Streams, but following Laravel's package development practices I ended up doing that.

In my package's service provider, I have those lines:
```php
public function register()
	{
            app('streams.addon.binder')->register('tag', 'warn', dirname(__FILE__).'/../../addons/tags/warn', 'Company\\Streams\\Addon\\Tag\\Warn\\WarnTag');
            app('events')->fire("streams::tags.registered");
	}
```


RyanThompson:
@nebulousGirl I got this implemented in a little differently as a core requirement.

PR #71, opened 2015-02-20
Allow alternative namespace(s) for Module installers

websemantics:
Changes to FieldInstaller, StreamInstaller and Addon classes (core) to enable a Module to install Fields and Streams in a namespace that's different than the module slug.

### DIFF
--- a/src/Addon/Addon.php
+++ b/src/Addon/Addon.php
@@ -88,9 +88,10 @@ class Addon implements PresentableInterface
      * @param  null $key
      * @return string
      */
-    public function getNamespace($key = null)
-    {
-        return "{$this->getVendor()}.{$this->getType()}.{$this->getSlug()}" . ($key ? '::' . $key : $key);
+    public function getNamespace($key = null, $namespace = null)
+    {   
+        $namespace = ($namespace)?$namespace : $this->getSlug();
+        return "{$this->getVendor()}.{$this->getType()}.{$namespace}" . ($key ? '::' . $key : $key);
     }
 
     /**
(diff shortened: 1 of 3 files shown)


websemantics:
I like the convention of using the module slug for namespace, but I'd like to break it when I have to :)

websemantics:
Here's an alternative to the getNamespace function
```
    public function getNamespace($key = null, $namespace = null)
    {
        return "{$this->getVendor()}.{$this->getType()}.{$this->getSlug()}" . ($namespace? ".{$namespace}" : "") .($key ? '::' . $key : $key);
    }
```

RyanThompson:
https://github.com/anomalylabs/streams-platform/commit/bfa70101034259f28dd65b81221ec1cd91b22a77

RyanThompson:
Looked like there was one more fallback needed.

The $namespace property in field installers will be a first default with fallback to the addon slug - priority is the field['namespace'] if it exists.

Stream installer just needs $stream to have a namespace in the array and it falls back to addon slug.


On Feb 20, 2015, at 4:00 PM, Adnan Sagar, PhD. <notifications@github.com<mailto:notifications@github.com>> wrote:


I like the convention of using the module slug for namespace, but I'd like to break it when I have to :)

websemantics:
Perfect, .. but would the current encoding of 'name', 'label', ' instructions' etc using getNamespace protects protects against name conflict across different namespaces in the same module?

Ex:
```
"vendor.module.example::field.language.name"
```

RyanThompson:
The vendor.type.slug namespace is strictly enforced - it’s not something you can override. The reason being is it’s basically a dot reference to your class namespace - but simpler to work with and predict. That has to stay.

On Feb 20, 2015, at 4:42 PM, Adnan Sagar, PhD. <notifications@github.com<mailto:notifications@github.com>> wrote:


Perfect, .. but would the current encoding of 'name', 'label', ' instructions' etc using getNamespace protects protects against name conflict across different namespaces in the same module?

Ex:

"vendor.module.example::field.language.name"

websemantics:
Thanks Ryan ..

PR #79, opened 2015-03-02
Websemantics patch 1

websemantics:
This would allow to override the default module slug as namespace.

### DIFF
--- a/src/Database/Migration/Command/Handler/MigrateFieldsHandler.php
+++ b/src/Database/Migration/Command/Handler/MigrateFieldsHandler.php
@@ -55,7 +55,7 @@ class MigrateFieldsHandler
 
             $field['slug']      = $slug;
             $field['type']      = array_get($field, 'type');
-            $field['namespace'] = array_get($field, 'namespace', $namespace ?: $addon ? $addon->getSlug() : null);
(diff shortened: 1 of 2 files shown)


RyanThompson:
I think what you are after is already accommodated here. 

The order of precedence is:

1.) The field namespace key
2.) The migration $namespace property
3.) The addon slug

websemantics:
True, .. but when I set the $namespace property in migration file, i.e. `protected $namespace = 'core';` it gets overridden by the addon slug!

RyanThompson:
I see - and this commit successfully allows both (the parenthesis make sense now.. Seems like Ive run into that before).

RyanThompson:
can you look through those migrate / roll back commands and see if there is any more instances of this logic that needs to be replaced?

websemantics:
For sure, .. I'm doing that right now ..

RyanThompson:
You da man

websemantics:
When uninstall a module, RollbackFieldsHandler leaves fields translation behind (should be deleted)

RyanThompson:
That should be triggered? https://github.com/anomalylabs/streams-platform/blob/1.0/develop/src/Model/Listener/DeleteTranslations.php

RyanThompson:
If you can't push a fix open a new issue mkay? ^_^

PR #106, opened 2015-06-02
Allow globbing of asset files

bloveless:
Granted asset_add can already glob a single directory using php glob function will allow for including nested directories, as well as any globbing magic that php allows. This might be able to replace the is_dir in the condition above it, but I was unsure if it would have any negative side effects by removing it from the line above.

### DIFF
--- a/src/Asset/Asset.php
+++ b/src/Asset/Asset.php
@@ -120,6 +120,13 @@ class Asset
         if (starts_with($file, ['http', '//']) || file_exists($file) || is_dir(trim($file, '*'))) {
             $this->collections[$collection][$file] = $filters;
         }
+        // if none of the previous conditions were met then lets check and see if globbing the file
+        // will return any results
+        else if (count(glob($file)) > 0) {
+            foreach(glob($file) as $curFile) {
+                $this->collections[$collection][$curFile] = $filters;
+            }
+        }
 
         if (
             config('app.debug')


RyanThompson:
You can use glob like this too:

`foo/bar/*/*` if you want deeper globing. As well as stuff like `foo/bar/*.js/*.min.js`

RyanThompson:
I think that functionality is a little more tightly controlled than your PR - thanks for the input though!

bloveless:
I did try doing something like foo/bar/*/* and it wasn't resolving any files. That was why I added the glob. Was I doing it wrong? It wasn't returning any files when I used {{ asset_add("header.js", "theme::js/remote/all/*/*", ["debug"]) }}

RyanThompson:
I think the glob detect in there before it hits assetic might be sucking..

Try changing the asset class on line 120 and 324 to something more like str_contains($file, '*')

RyanThompson:
Ideally we'll have a preg_match so it can handle more complex matches..

bloveless:
I think I see what you are saying. I'll take a look and see what I can find out.

RyanThompson:
Your `count(glob("$dir/*")` might actually be the easiest solution preg_matching could get messy:

http://www.cowburn.info/2010/04/30/glob-patterns/

In add can you use your count glob instead of the * crap and add a "glob" filter if it is glob (second IF statement not elseif.

Then in the 324 do if glob is in filters then use glob class.

RyanThompson:
See if that get's you intended results and resubmit if you would - nice work! Never thought of this stuff the assets class was of the first I wrote.

PR #107, opened 2015-06-04
Lets try asset globbing once more

bloveless:
Alright I think I hit the points we were talking about in the other pull request. This guy adds the glob filter and then actually uses the GlobAsset class found later in the class. Tested against foo/bar/*/* and it is working.


RyanThompson:
Sick nasty - nice work. Ima merge and clean up commenting a bit cause Im OCD evidently ;-)

PR #112, opened 2015-06-08
Fix tree children with intval on getParent method of Item class

bloveless:
Found the bug that was not showing children in the item tree. There will be this pull request and another. Choose one or the other, but really it would work if you accepted both of them. This one ensures that the getParent method of the item class is indeed an integer so that the === in the children function of ItemCollection will be true.

### DIFF
--- a/src/Ui/Tree/Component/Item/Item.php
+++ b/src/Ui/Tree/Component/Item/Item.php
@@ -95,7 +95,7 @@ class Item implements ItemInterface
      */
     public function getParent()
     {
-        return $this->parent;
+        return intval($this->parent);
     }
 
     /**


RyanThompson:
This fixes an issue somewhere too - Ill try and find it. Weird how it works fine for me but some environments it doesn't.

Nice find!

bloveless:
Just out of curiosity what version of PHP are you using? That is the only thing I can think of. I'm using 5.5.24

RyanThompson:
5.5.10 and it works for me. I believe the gentleman that brought it to my attention prior was on Windows / XAMP.


On Jun 8, 2015, at 2:48 PM, bloveless <notifications@github.com<mailto:notifications@github.com>> wrote:


Just out of curiosity what version of PHP are you using? That is the only thing I can think of. I'm using 5.5.24

bloveless:
Very strange.

PR #113, opened 2015-06-08
Fix tree children by using a non exact equality. == rather than ===

bloveless:
Found the bug that was not showing children in the item tree. There will be this pull request and another. Choose one or the other, but really it would work if you accepted both of them. This one uses == rather than === so that the string "2" will equal the int(2) as is returned from the getParent function.

### DIFF
--- a/src/Ui/Tree/Component/Item/ItemCollection.php
+++ b/src/Ui/Tree/Component/Item/ItemCollection.php
@@ -45,7 +45,7 @@ class ItemCollection extends Collection
 
         /* @var ItemInterface $item */
         foreach ($this->items as $item) {
-            if ($item->getParent() === $parent->getId()) {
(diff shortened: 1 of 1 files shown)


RyanThompson:
I assume this is in conjunction with the last one regarding intval?

bloveless:
Yeah. I found two ways to solve the problem and I was hoping that you would choose one and close the other. Just depending on which PR you liked better. Or both... really they will work together if you wanted to use them both.

RyanThompson:
Ill pry remove the intval then if this works on it's own. Cleaner. I don't use that anywhere else so it makes me itch a bit seeing it knowing it's not needed lol..

bloveless:
Cool cool. And yes this commit will work on it's own.

PR #117, opened 2015-06-29
Use rawurlencode rather than urlencode

bloveless:
http://stackoverflow.com/questions/996139/urlencode-vs-rawurlencode

I was still having some issues creating sub folders of folders that had spaces. I was able to create " Main Content Images > Home Page > Test Folder > This is a test folder! > And another test folder > Last Test" using this function.


RyanThompson:
Works for me!

PR #118, opened 2015-06-30
Seems like the router doesn't like %2F

bloveless:
I don't know why it was working yesterday, but after a fresh install today something doesn't like %2F and works when they are a /...

Just to be sure I've deleted everything, pulled, composer update, verified issue, applied fix, and tried it again.

I don't feel the greatest about this monkey patch, but figured it would be better than just posting an issue.

### DIFF
--- a/src/Support/Parser.php
+++ b/src/Support/Parser.php
@@ -102,7 +102,7 @@ class Parser
                 'route'   => [
                     'parameters'            => array_map(
                         function ($value) {
-                            return rawurlencode($value);
+                            return str_replace('%2F', '/', rawurlencode($value));
                         },
                         $route->parameters()
                     ),
@@ -110,7 +110,7 @@ class Parser
                         '/',
                         array_map(
                             function ($value) {
-                                return rawurlencode($value);
+                                return str_replace('%2F', '/', rawurlencode($value));
                             },
                             $route->parameters()
                         )


RyanThompson:
Where is a slash coming from? In your folder name?

bloveless:
It is coming from a sub folder. The top folder is Main Content Images and the sub folder is Home Page. The slash is Main Content Images/Home Page (Main%20Content%20Images**%2F**Home%20Page)

bloveless:
I haven't tried adding a slash to the folder name. I assume that will have an issue as well.

RyanThompson:
Can you change the raw encode to the regular encode and try again.

Also ?_publish and clear browser cache just in case

bloveless:
screen shot 2015-06-30 at 5 37 50 pm

screen shot 2015-06-30 at 5 38 37 pm

That is using the regular urlencode after doing a ?_publish

bloveless:
It is 404'ing on the modal popup.

RyanThompson:
That slash should not be encoded...

RyanThompson:
Let me take a lookie here..

RyanThompson:
Ah it's cause $path in the parameters is a single parameter - so it will contain a slash in nested folders..

RyanThompson:
This class has issues..

Here's what's causing it - the parameters in my case were:

array:2 [▼
  "disk" => "test_local"
  "path" => "Test Folder/Test Again"
]

So the parameter is the path and we need to preserve those slashes.. 

I think the root of the issue is that I was being lazy with adding this in the first place.. it needs more thought. I am rebuilding. Will push shortly.

RyanThompson:
Fixed: https://github.com/anomalylabs/streams-platform/commit/c1d9455bc0ca1691fe9991036f8c2c90adb5247a

Issue #119, opened 2015-07-02
Default order_by if sortable equals true


bloveless:
If sortable is equal to true then even if order_by is not set then set order_by

        'order_by' => [
            'sort_order' => 'ASC'
        ],

RyanThompson:
Donezo: https://github.com/anomalylabs/streams-platform/commit/3a375005f34f7f59326f40c7edb810bc0a126144

bloveless:
So this is weird. Sortable is equal to true. When I dd the table at the end of the \Table\Command\SetDefaultOptions.php I see that the table options are set to order_by => [ 'sort_order' => 'asc' ]

![screen shot 2015-07-02 at 2 56 53 pm](https://cloud.githubusercontent.com/assets/535135/8487636/97a9a452-20ca-11e5-951e-addc0b816de7.png)

But the query in the page is still show sort by id

![screen shot 2015-07-02 at 2 57 12 pm](https://cloud.githubusercontent.com/assets/535135/8487646/a55036fc-20ca-11e5-9f8a-fc9d998a605d.png)

RyanThompson:
Anything weird in here: 

https://github.com/anomalylabs/streams-platform/blob/1.0/develop/src/Ui/Table/Command/SetDefaultOptions.php#L108

bloveless:
Nope it looks exactly the same as what you've got.

PR #120, opened 2015-07-02
If the model is both EntryModel and EloquentModel don't clobber over …

bloveless:
…each other

https://github.com/anomalylabs/streams-platform/issues/119


RyanThompson:
Indeed! The entry model directly extends the eloquent model. Good catch.

RyanThompson:
Yeeeep

PR #121, opened 2015-07-02
Same thing different line

bloveless:


RyanThompson:
Just a heads up I don't think this is PSR. else if should be elseif

Ill adjust I usually use PHPStorm to mass format anything I miss.

bloveless:
I'll get it next time. Do you use the default phpstorm php code formatting? Or some custom styles?

PR #129, opened 2015-07-10
Always put the buttons on the same line in tables

bloveless:


RyanThompson:
Nice! I didn't actually know about that tag :-/

PR #136, opened 2015-07-23
Adding StylusFilter

websemantics:


RyanThompson:
I can dig it. Just need to doc the dependency.

websemantics:
Yes, dependency should be fed, possibly from config or .env

PR #138, opened 2015-07-23
Use filemtime to determine if images need to be published to file system

bloveless:


RyanThompson:
Sexy

RyanThompson:
Much better than my implementation lol

PR #147, opened 2015-08-24
german translation, first round

marcogrueter:
first round of german translations - expect some future changes, depending on the needs of the other parts/repos/modules and because I don't have a "expression road map" or master plan (yet)


RyanThompson:
You rock!

PR #152, opened 2015-09-14
Fix ExceptionHandler for when 'app.debug' is false

JoelESvensson:
The returned ResponseFactory from response() doesn't have a setStatusCode method. Use view($template, $vars, $statusCode) instead


RyanThompson:
Thank you kindly!

PR #154, opened 2015-09-17
psr2 fixes (fixes #151)

### LINKED ISSUE #151: add .php_cs or phpcs.xml
Hello,

I think it would be good idea to add 'phpcs' to the repo..
I see a lot PSR2 violations like:

"if (!$var)" should be "if (! $var)"

I think "initial" changes should be made by you, @RyanThompson, what do you think?

if you would like to use Fabiens php-cs-fixer (cs.sensiolabs.org), then config should be taken from here: https://raw.githubusercontent.com/laravel/framework/5.1/.php_cs

and just run: `php-cs-fixer fix`

Because the more far we go, we more those changes will be later

THanks

huglester:
Hello,

what do you think about this, @RyanThompson ?

If it's ok, I think I'd go all over packages to add both .editorconfig + .php_cs

Thanks


RyanThompson:
Are you sure these are ALL PSR2 and not Symfony CS in addition? I know they have their own CS style in addition to PSR2 standards.

RyanThompson:
PS love the idea of this - something to smack down the LAW on what is right and wrong (something without a human soul at that lol)

huglester:
Hello, 

I thought I answered this, seems no.
Well PSR1 was to use ie:
```php
if (!<space here>$app) {
}
```

PSR2:
There MUST be one blank line after the namespace declaration, and there MUST be one blank line after the block of use declarations.

I think all of those were/are valid.

I think if you are +1 for this I could re-create the pull request or you can take the config php .php_cs and run it yourself (the one config in this PR). So you can take a closer look on what was/is happening

PR #156, opened 2015-09-23
Default value not supported on TEXT column

tomcastleman:


RyanThompson:
Ahh beat ya to it. Good looking out though thank you!

PR #159, opened 2015-10-14
Add missing symfony/yaml dependency

PMLavigne:
Add dependency on symfony/yaml ~2.0, equivalent to the version used by phpunit. Yaml was only being pulled in by dev dependencies, therefore using composer's --no-dev left the project without it. 

Closes pyrocms/pyrocms#3550


RyanThompson:
Thank you!

PR #164, opened 2015-11-14
Update the code name for Chinese.

jacksun101:


jacksun101:
All chinese traditional translation for streams platform is done.

RyanThompson:
You rock! Thanks so much.

PR #174, opened 2016-01-16
Update Image.php

ramcda:
Must be able to properly set orientation on images from mobile devices.  This has been thoroughly tested


RyanThompson:
Thank you!

PR #176, opened 2016-02-05
Update LoadCurrentTheme.php

ramcda:
Fix issue with overrides being active simultaneously from both the admin and standard theme


RyanThompson:
Pushed fix per our discussion ^_^

PR #184, opened 2016-03-30
fix missing ORIGINAL_REQUEST_URI index

keevitaja:
$_SERVER['ORIGINAL_REQUEST_URI'] is not defined when borwsing the root URI.

This pull request fixes it!


RyanThompson:
Thanks!

PR #190, opened 2016-05-03
Cannot use getimagesize on images from files module since php doesn\'…

bloveless:
…t know where those files are

### DIFF
--- a/src/Image/Image.php
+++ b/src/Image/Image.php
@@ -745,11 +745,6 @@ class Image
         if (is_string($image) && str_is('*://*', $image) && !starts_with($image, ['http', 'https'])) {
 
             $this->setExtension(pathinfo($image, PATHINFO_EXTENSION));
-
-            $size = getimagesize($image);
-
-            $this->setWidth(array_get($size, 0));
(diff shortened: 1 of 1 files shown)


RyanThompson:
This isn't really a fix we need height / width

bloveless:
There needs to be something similar to $image = $this->paths->realPath($image); from the block above it, but I'm not sure what it should be. I did try the realPath function there, but it only works if the string has a '::' in it.

RyanThompson:
I'll check on a puter - if it's just that string identifier maybe this is ok.. Or we can load the resource or something

bloveless:
That particular section seems to only deal with string identifiers that don't start with http or https. I.E. local:// and it barfs when it gets one of those. I suppose it only throws a warning, but warnings = errors

RyanThompson:
Try loading it with the mount manager / it's done elsewhere in that very same class

bloveless:
I can't imagine that this is what you are looking for, but hey... it's really late, and I made the bastard work.

RyanThompson:
Just ended up taking that out - not supported there

PR #191, opened 2016-05-06
Set Form Entry before Options and Default Options

bloveless:
Setting the form entry before the setting the form options and default form options allows us to use the entry data in the form options. I.E. using the entry name (if it exists) in the form title.


RyanThompson:
I can dig this - but, let's set it as soon as possible. Go!

bloveless:
Alright so the order that made sense to me was SetFormModel -> SetFormStream -> Set Repository -> SetFormEntry -> then the rest

bloveless:
@RyanThompson What do you think about this?

RyanThompson:
Perfect! And if you need it earlier then you need to set it yourself ^_^

PR #200, opened 2016-05-24
Default the table limit to 9999 when the table is sortable

bloveless:


RyanThompson:
Beat you to this one!

RyanThompson:
Actually your's is slightly different.. merging in.

PR #211, opened 2016-05-27
1.0/develop

RyanThompson:


RyanThompson:
This "properly" fixes the relation method issue during install and preserves behavior in the control panel.

PR #216, opened 2016-06-02
Fields should allow labels to be hidden

bloveless:


resources/views/form/partials/wrapper.twig
```diff
@@ -8,7 +8,7 @@
             {{ field_type.hidden ? 'hidden' }}
             ">
 
-    {% if field_type.config.disable_label == true %}
+    {% if field_type.config.disable_label != true %}
```
RyanThompson: Geez la'peetz

PR #221, opened 2016-06-07
Bug table options should accept strings

bloveless:
When an options handler is used it will be set as a string. setOptions needs to accept both strings and arrays.


bloveless:
Not ready yet.

PR #222, opened 2016-06-07
Bug table options should accept strings

bloveless:
Okay this one works all the time. We have to be able to come from a string handler and go to an options array, add just an options array, or merge options together to match the functionality from before.


RyanThompson:
I feel like this should just drop the options array hint and drop the merge too.. Maybe add an addOption method or something. But this much logic in a setter is weird

bloveless:
Yeah it felt weird to me too. I didn't know if you had the array_merge in there for a specific reason or not.

PR #229, opened 2016-06-10
Reorder should allow sort_order to be updated by entry_id and related…

bloveless:
…_id since many-to-many tables don't have ID's


bloveless:
New way is WAY better than this!

PR #233, opened 2016-06-14
add and fix something

rayyang88:


RyanThompson:
There were other commits in here but for the language addition - so I pulled from your fork and merged manually! Thanks for the translations ^_^

rayyang88:
@RyanThompson how about the other fix? this commits https://github.com/anomalylabs/streams-platform/pull/233/commits/230636ba03a00b7bdf97fa793fb516796b886a67
i want to customize the buttons.

RyanThompson:
Can you explain exactly what you are trying to do? I am not quite following you Im sorry

rayyang88:
@RyanThompson 
Sorry for replying so late.
The commit fix two bugs:
1. If I change the app language to zh-CN, some fields of some model will disappear, because the code uses app.fallback_locale, not the default locale
2. If I want to customize the restore button or force_delete button, and I add some configs to the button like this
image
then it didn't work, so I fix the bug, then it works fine now:)

RyanThompson:
Sorry this PR still needs to be fixed up. There are translations files along with a fix in here. I need separate PRs for each concern please!

PR #260, opened 2016-07-05
Correctly count force deleted entries

bloveless:
Since the force delete can be caught and rejected by an observer, we should check that the entry was actually deleted rather than returning true.

I noticed this when I tried to force delete and entry, but rejected it due to another constraint, but the table still showed 1 row(s) successfully deleted.

screen shot 2016-07-05 at 11 28 13 am


RyanThompson:
Submit to `develop` please ^_^

bloveless:
Damn it all! I need to be able to set develop as the default branch!

PR #264, opened 2016-07-11
Allow the use of old in twig.

bloveless:


RyanThompson:
`{{ request_old(key, default) }}` should work

RyanThompson:
https://github.com/anomalylabs/streams-platform/blob/master/src/StreamsPlugin.php#L287

bloveless:
Whoooooops! That worked :) Thanks!

RyanThompson:
^_^

PR #269, opened 2016-08-05
Image __toString should always return a string

bloveless:
Even if there is no output (missing image or error) the image __toString should alwasy return a string.

I just got tired of missing or errored images throwing a 500. What do you think about just returning an empty string if the output isn't already a string?

    ErrorException: Method Anomaly\Streams\Platform\Image\Image::__toString() must return a string value in /var/www/app.findingtruth.io/public_html/vendor/twig/twig/lib/Twig/Extension/Core.php:968

### DIFF
--- a/src/Image/Image.php
+++ b/src/Image/Image.php
@@ -1144,7 +1144,12 @@ class Image
      */
     public function __toString()
     {
-        return $this->output();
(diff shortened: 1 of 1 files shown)


RyanThompson:
I dig it

PR #270, opened 2016-08-07
Show the admin status in the js constants.

bloveless:
I was thinking it would be nice to let the front-end js change if you want to add some features when the user is logged into the admin.

### DIFF
--- a/resources/views/partials/constants.twig
+++ b/resources/views/partials/constants.twig
@@ -12,4 +12,6 @@
     var TIMEZONE = "{{ config_get('app.timezone') }}";
     var LOCALE = "{{ config_get('app.locale') }}";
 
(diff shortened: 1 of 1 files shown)


RyanThompson:
You should be able to 'user().admin' but is this necessary? You can include this stuff in assets.

bloveless:
Not necessary. I just found it useful today and thought I would propose it. I like to add things for front end admin users to make their lives easier and thought it would make sense in there. You can close this if you disagree.

RyanThompson:
I wouldn’t say that I disagree necessarily. But needs a good definition / scope of what we need to do.

Normally I would add data-* attributes for the stuff I need and use it the way I need. I haven’t found that a normal set of principles can be resolved from the needs I’ve ran into just yet.. 

Ill close it for now.

Ryan

PR #276, opened 2016-08-31
Remove attr disabled on drag drop of table row

bloveless:
We also have to remove the attr disabled on the table button when a row is dragged and dropped otherwise the button will still not be clickable, but will required a checkbox to be checked to enable the reorder button.


RyanThompson:
Can you move this to master please. Sorry, but text me real quick and Ill show you the git flow.

PR #280, opened 2016-09-10
localization: adding ar folder to Lang

aldf:


aldf:
this necessary because otherwise when changing the language from Pyrocms Admin footer  it will display   only the translation and cache the old scss files that mean no changing to RTL or LTR

RyanThompson:
Please undo the changes to the application so that this is ONLY a localization change. I have an easier method I will implement when I merge in the RTL change. No need to clear assets.

aldf:
OK , done.

RyanThompson:
Fantastic - thank you!

PR #287, opened 2016-09-30
Update scssphp

mikey0000:
allow minor version bumps


RyanThompson:
Has your fix been merged? This was purposeful but I believe the original issue was resolved since.

mikey0000:
No the original issue I believe has not been fixed and may not get fixed. I am however waiting on a couple of other PR's so node sass support can be added.

PR #291, opened 2016-10-04
Maybe use App\Http\Middleware\VerifyCsrfToken?

huglester:
Hello,
maybe better to use: $this->middleware('App\Http\Middleware\VerifyCsrfToken'); for easier extendability?
I know we can override via IOC, but this is more 'laravel' intuitive? Just my two cents :)

You would need to override $except = [] often

### DIFF
--- a/src/Http/Controller/BaseController.php
+++ b/src/Http/Controller/BaseController.php
@@ -131,7 +131,7 @@ class BaseController extends Controller
         }
 
         // These may be manipulated by the middleware above.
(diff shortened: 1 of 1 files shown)


RyanThompson:
We can't assume `App` because often times developers change the app name. 

You can still override the `\Illuminate\Foundation\Http\Middleware\VerifyCsrfToken` class by binding your own to it early on. Try defining the binding in `config/streams.php@middleware`.

PR #295, opened 2016-10-19
Adds support for handling options with a callback

aliselcuk:


src/Ui/Table/Component/Filter/Type/SelectFilter.php
```diff
@@ -58,6 +58,9 @@ public function getOptions()
      */
     public function setOptions(array $options)
     {
+        if (array_key_exists('handler', $options)) {
+            $options = app()->call($options['handler']);
+        }
```
RyanThompson: 1.) Use the Resolver (see docs -> services)
2.) Do it in the `getInput` method and pass the filter of course. So the handler can do `$filter->setOptions($value)`

PR #297, opened 2016-10-23
Fix Field type [] not found.

huglester:
Now we get this:

[Exception]
  Field type [anomaly.field_type.video] not found.


RyanThompson:
Good call!

PR #301, opened 2016-10-31
Adds option for injecting raw html into section tabs

aliselcuk:
should we also do this for normal sections? (non-tabbed ones)


RyanThompson:
Can you add HTML to the non-tabbed sections as well please? I dig it.

RyanThompson:
And add support for tab.view as well if you would.

RyanThompson:
Excellent job - thanks!

PR #305, opened 2016-11-02
Replaced logical operators

Newman101:
Hi. I spotted couple of logical operators (namely, ```or``` and ```and```) in the source code. 

The problem with these operators is that they do not have the same precedence as ```||``` and ```&&```. This could lead to unexpected behavior. 

This issue has now been amended.

Thank you.


RyanThompson:
Thanks! Good call - not sure how those slipped in there ^_^

PR #307, opened 2016-11-03
Amended permissions for zh-cn language files

Newman101:
Hi. 

Looks like Chinese language pack included files which were executable.

This is a major security hazard due to too permissive permissions. 

The issue has been now amended. 

Thank you.


RyanThompson:
Thank you!

PR #308, opened 2016-11-03
Add a runtime query cache

bloveless:
for any models that are being queried more than once.

### DIFF
--- a/src/Model/EloquentQueryBuilder.php
+++ b/src/Model/EloquentQueryBuilder.php
@@ -40,7 +40,7 @@ class EloquentQueryBuilder extends Builder
     {
         $cacheKey = $this->getCacheKey();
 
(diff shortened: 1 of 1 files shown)


src/Model/EloquentQueryBuilder.php
```diff
@@ -24,34 +24,51 @@ class EloquentQueryBuilder extends Builder
     protected $model;
 
     /**
+     * Store a runtime cache so that there are never any duplicated queries.
+     *
+     * @var array
+     */
+    protected static $modelCache = [];
```
RyanThompson: Lets just call it cache - you know how I feel about multiple syllables and words.

bloveless: LOL you got it!

bloveless:
Updated to a single word :)

PR #314, opened 2016-11-04
Fix pgsql driver slug

Piterden:
In ConfigureDatabase.php file you've used this command to configure DB:
```php
require base_path('config/database.php')
```
But there is'n `postgres` driver in `database.php` in Laravel's config, there is only `pgsql`


RyanThompson:
Good call thank you!

Piterden:
Not so good anyway. I wasn't able to make correct install CMS with pgsql driver.
First problem was duplicated indexes (pg don't want the same naming indexes even in different tables)
I make much fixes in migrations of core and modules, but this wasn't help me too.

PR #321, opened 2016-11-16
Adds firing event after form is validated

aliselcuk:
I really needed this on the FormBuilder@onValidated. Someone might need too.

Was it forgotten or dropped intentionally?


aliselcuk:
use form handler instead

PR #324, opened 2016-12-01
Added newline character to several files

Newman101:
Hello.

Couple of fixes: 

1. Several files were missing the newline character in the end of file. This has a potential to cause the file to render incorrectly on a terminal, and it's considered an anti-pattern in general. The files are now fixed. 

2. There was ```-/nbproject/private/``` artifact in .gitignore file. This was most probably accidentally committed by NetBeans, The artifact has been removed. 

Thank you.


RyanThompson:
Fantastic - thanks as always!

PR #326, opened 2016-12-03
Added missing exception handling in Kernel.php

Newman101:
Hi.

Looks like there was a bit risky use of ```parse_url``` function in src/Http/Kernel.php. 

Basically: parse_url can return boolean false if the URL is malformed. Should this happen, the hackers will be able to pass seriously malformed URLs to perform hostile XSS injections. 

I created an if-clause for malformed URLs. Basically, if variable ```$url``` gets malformed URL which makes the parse_url to return false, the Kernel.php throws an exception. 

Thank you.


RyanThompson:
Outstanding.

PR #330, opened 2016-12-13
Allow forms to persist between submits in case there are any errors.

bloveless:
We do this by not forgetting the cache unless the builder was submitted
without errors.

This introduced another issue. Where the errors were being applied to
the form again via the session. To solve this I reset the form errors
when the form is being validated. This allows the old errors to be
forgotten and the new errors to persist through.

### DIFF
--- a/src/Http/Controller/FormController.php
+++ b/src/Http/Controller/FormController.php
@@ -40,7 +40,10 @@ class FormController extends PublicController
             ->getFormResponse();
 
         $builder->flash();
-        $cache->forget('form::' . $key);
+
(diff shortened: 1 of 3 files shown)


RyanThompson:
Merged this in manually cause there where a few issues I had with it - just small shit - thanks for the solution!

RyanThompson:
https://github.com/anomalylabs/streams-platform/commit/e6155d04c05555fc8465cfe869b196b7336d37b5

bloveless:
Ha ha I didn't even realize that the form builder had a set errors function. As long as it is in there I'm happy :)

PR #332, opened 2016-12-14
Replaced non-existend method in EloquentRepository.php

Newman101:
Hi.

Method ```findTrashed``` was calling ```onlyTrashed``` method in EloquentModel. How ever, this method does not exist. 

I've replaced this method call with the correct one. 

Thank you.


RyanThompson:
Thank you much!

PR #335, opened 2016-12-17
Added merge skips to FormBuilder

daviesgeek:
Used exactly like `mergeOptions`:

```php
$builder->mergeSkips(['field_1', 'field_2]);
```


RyanThompson:
I am merging this in manually cause of CS. Be sure to run your PSR code beautifiers with PHPStorm before committing ;-)

PR #338, opened 2016-12-30
Added the ability to install bower and npm on module install

daviesgeek:
Test PR for https://github.com/pyrocms/pyrocms/issues/4027

### DIFF
--- a/src/Addon/Module/Command/InstallModule.php
+++ b/src/Addon/Module/Command/InstallModule.php
@@ -69,6 +69,8 @@ class InstallModule
 
         $modules->install($this->module);
 
+        $modules->installNpm($this->module);
+
         $manager->register();
 
         if ($this->seed) {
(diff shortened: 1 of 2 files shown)


daviesgeek:
@RyanThompson here's what I have so far. I'm not sure if this is too naive of a solution?

daviesgeek:
A couple of missing things:

This also needs to be implemented for extensions, not just modules. I didn't add this for extensions, as I wanted to see if this was on the right path.

Field types are not installed, and thus, I'm not really sure exactly how to get the dependencies installed. I'll think about this and see if I can come up with a good plan for it.

src/Addon/Module/Command/InstallModule.php
```diff
@@ -69,6 +69,10 @@ public function handle(
 
         $modules->install($this->module);
 
+        $modules->install($this->module);
```
RyanThompson: This is duplicated?

RyanThompson:
Also does this install it within the addon itself or in the project root?

daviesgeek:
It installs within the addon; not the root. That way you can reference it relative to the addon's root within your views or whatever needs it.

daviesgeek:
Added support for using yarn if there's a `yarn.lock` file in the root of the project, and yarn is installed.

RyanThompson:
Sorry man - Ima sit on this for a little while.

PR #341, opened 2017-01-02
Fixed a bug with the field filler removing the first field

daviesgeek:
This happens if you set the fields for a form builder like this:

```php

$builder->setFields([
  '*',
  'other_field' => 'anomaly.field_type.text'
]);

```

And where the form builder has a stream entry with some fields like this:

```php
$fields = [
  'title',
  'slug' => [
    'config' => [
      'slugify' => 'title'
    ]
  ]
];
```

When the `FieldFiller` loops over the fields, it's not checking to make sure the `array_search` actually returns something (and not `false`), so I added a check to make sure that it doesn't return false. Otherwise the first field will get removed (title).

### DIFF
--- a/src/Ui/Form/Component/Field/FieldFiller.php
+++ b/src/Ui/Form/Component/Field/FieldFiller.php
@@ -52,7 +52,13 @@ class FieldFiller
                 continue;
             }
 
-            unset($fill[array_search($parameters['field'], $fill)]);
+            $search = array_search($parameters['field'], $fill);
+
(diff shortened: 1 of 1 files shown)


RyanThompson:
Merged this in manually cause of CS issues.

Please run a PSR-2 formatter on it ^_^

daviesgeek:
I ran a formatter on it, so if needed, this PR can be merged now :)

RyanThompson:
Sorry I should have closed this - I merged it in manually.

PR #343, opened 2017-01-07
Allow sections to be hidden from the Control Panel.

bloveless:
This will allow sections to be set as hidden from the sections array. The control panel theme will have to be updated to respect this flag.

### DIFF
--- a/src/Ui/ControlPanel/Component/Section/Section.php
+++ b/src/Ui/ControlPanel/Component/Section/Section.php
@@ -112,6 +112,12 @@ class Section implements SectionInterface
     protected $breadcrumb = null;
 
     /**
+     * If the section will be hidden from the control panel.
+     * @var bool
+     */
+    protected $hidden = false;
+
+    /**
      * Get the slug.
      *
      * @return null|string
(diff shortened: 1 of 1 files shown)


bloveless:
For a theme example on how see this PR https://github.com/pyrocms/accelerant-theme/pull/5

src/Ui/ControlPanel/Component/Section/Section.php
```diff
@@ -435,6 +441,29 @@ public function setBreadcrumb($breadcrumb)
     }
 
     /**
+     * Get if this section is hidden.
+     *
+     * @return bool
+     */
+    public function getHidden()
```
RyanThompson: Can we do `isHidden` here ^_^

RyanThompson:
How would you designate this section as being active then? No sections would be marked active?

bloveless:
If there is a parent of a active child then it get a highlighted attribute. In this https://github.com/pyrocms/accelerant-theme/blob/4ebbc5730b2d67601a1726259bcf7cbea79960a1/resources/views/macros/menu.twig I'm also adding the class highlighted which would be applied to all the parent even if the child is hidden. I assume that is what you are asking about.

PR #344, opened 2017-01-11
Add South African Provinces to State field type

MikhailFerreira:


.gitignore
```diff
@@ -2,3 +2,4 @@ bin
 build
 vendor
 composer.lock
+.idea
```
MikhailFerreira: No problem - thanks for the feedback. :)

PR #345, opened 2017-01-14
Added skip uninstall addons if not installed

Piterden:


RyanThompson:
Is there any reason why we need to warn them and halt it? I can't think of any reason to allow it but I think it's a more graceful way of handling should there be anything left lying around for some reason.

Piterden:
The reason is in the `reinstall` artisan command.
When I setup database and "playing" with DB structure, I use `reinstall` command, which throws me an exception when my module is not installed (for some reasons).
This fixes it.

RyanThompson:
Ok fair enough - can you send me an example just so I can make sure there’s nothing more sinister wrong with it :-P

Ill get this merged in though.

Piterden:
Sorry, I don't understand, example of what?

Piterden:
It seems, that was fixed already.

PR #347, opened 2017-02-07
Added .eslintrc.yml config

Piterden:
If you allow)

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -8,6 +8,6 @@ indent_style = space
 indent_size = 4
 trim_trailing_whitespace = true
 
-[*.{yml,js}]
+[*.yml]
 indent_style = space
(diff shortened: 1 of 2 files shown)


RyanThompson:
What is this for? And I think we should ALWAYS indent with 4 spaces.

Piterden:
.eslintrc is for configure linter.
JS is really more convenient with 2, but if you don't want, this is your right.

RyanThompson:
I feel like this might be too personalized to be included.. thoughts?

RyanThompson:
I have to close this one I do believe. IMO this reflects a _personal_ project file. I personally use my IDE to lint. Perhaps later on we'll rely on an auto CS service like Laravel does.

Issue #348, opened 2017-02-08
[BUG] FiresCallbacks Trait is storing callbacks in wrong trigger keys


agustindidiego:
This line of code should be removed:
```php
$trigger = get_class($this) . $trigger;
```
It can be found in Trait Anomaly\Streams\Platform\Traits\FiresCallbacks.php line 54

PR #349, opened 2017-02-08
[BUG] FiresCallbacks Trait is storing callbacks in wrong trigger keys…

agustindidiego:
… #348

 - Removing unneded listener key naming

### DIFF
--- a/src/Traits/FiresCallbacks.php
+++ b/src/Traits/FiresCallbacks.php
@@ -51,8 +51,6 @@ trait FiresCallbacks
      */
     public function listen($trigger, $callback)
     {
-        $trigger = get_class($this) . $trigger;
-
         if (!isset(self::$listeners[$trigger])) {
             self::$listeners[$trigger] = [];
         }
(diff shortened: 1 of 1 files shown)


agustindidiego:
I fixed and improved the way the listeners are being searched and called.
As an example of usage I'll reference the FormBuilder class.
With this fix you are able to listen to events fired on any Form that extends FormBuilder and/or listen to events fired on a extended FormBuilder class.
Example:
Having $formBuilder(FormBuilder), $my1formBuilder(MyFirstFormBuilder) and $my2formBuilder(MySecondFormBuilder)...

```php
$formBuilder->listen('ready', function($builder){
    // This listener is being called 2 times, once for each Form
    // $builder will be an instance of MyFirstFormBuilder on one call
    // and an instance of MySecondFormBuilder or another
});

$my1formBuilder->listen('ready', function($builder){
    // This listener is being called 1 time
    // $builder will be an instance of MyFirstFormBuilder on one call
});

// I can add a second listener to $my1formBuilder on the same event
$my1formBuilder->listen('ready', function($builder){
    // This listener is being called 1 time
    // $builder will be an instance of MyFirstFormBuilder on one call
});
```

RyanThompson:
Fantastic. Very nicely done! Thank you ^_^

PR #350, opened 2017-02-09
fix currency_format

huglester:
With old formating in place, you will have ambigous numbers.
not sure why it is needed at all here.

we format value, add prefix/suffix and thats it!

thanks

### DIFF
--- a/src/Support/Currency.php
+++ b/src/Support/Currency.php
@@ -71,7 +71,7 @@ class Currency
             $suffix = $this->symbol($currency);
         }
 
-        return $prefix . number_format(floor(($number * 100)) / 100, $decimals, $point, $separator) . $suffix;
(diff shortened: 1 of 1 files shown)


RyanThompson:
Can you explain "ambigous numbers"?

The purpose is to take 10.013 and convert to 10.01 and 10.377 to 10.38

RyanThompson:
I am going to close this for now and wait to get some heavy real world testing. I don't like the string thing (also it returns a string) also. Might revisit but I need more experience with it / proof of concepts for the two approaches.

stefanpkc:
I don't want to reopen this, but just for when this takes off again, there's an argument not to use floats or decimals at all. Only for presenting/printing out values you divide by 100 to get the decimals.

PR #352, opened 2017-02-10
Rows (when using columns) doesn't have the correct styles

bloveless:
Wrapping the rows in an additional .field-group pulls in the correct styles for the rows.

I'm also pushing into master since develop doesn't have the rows form file.

Before:

screen shot 2017-02-10 at 12 49 15 pm

After:

screen shot 2017-02-10 at 12 49 36 pm


RyanThompson:
Thanks!

bloveless:
No Worries!

PR #353, opened 2017-02-13
Give more helpfull error when field not existing

huglester:
Hello,

I had this in migrations:
```php
'is_hidden' => [
            'type' => 'boolean',
        ],
```
after some time, this field type was just not found. I remember it happening in other module also.

Thanks!


RyanThompson:
This is already done ^_^

PR #354, opened 2017-02-13
More convenient exception

Piterden:
Showing which of fields exactly thrown error


RyanThompson:
Merged this in / expanded on it manually. THanks!

PR #358, opened 2017-03-02
Fix `admin/login` handling for logged in user

Piterden:
https://github.com/anomalylabs/users-module/pull/46


RyanThompson:
We'll just use first - we don't wanna force dashboard as the home ^_^

PR #359, opened 2017-03-10
Create search:rebuild artisan command

agustindidiego:


RyanThompson:
Hmm.. I have to remove this - this is doable now with scout: commands. I thought this would rebuild all streams. 

I am going to stash this code and put it into the upcoming search suite.

Thanks!

agustindidiego:
mmm, not quite..., scout:import command needs a single Model class as parameter, and it will only import one single Model at a time.
search:rebuild command can accept a single namespace::stream parameter, or just a namespace parameter. If a namespace is given without a stream, it will search and import all the streams that belongs to that namespace.

PR #362, opened 2017-03-14
Added `bare` option, which disables theme resources scaffolding.

Piterden:


RyanThompson:
I have other plans here that are similar ^_^

Piterden:
What are those other similar plans?

RyanThompson:
Driver based scaffolding / builders.

PR #365, opened 2017-03-21
Added `--force` option

Piterden:
Which is toggle `INSTALLED` .env flag at install time.


RyanThompson:
This doesn't do much because the app is already booted by this point.

RyanThompson:
better to use php artisan env:set INSTALLED=false / compile / redo installed

Piterden:
More intuitive would be flag `HELP_ME=true` to `.env` for that situation ^

PR #375, opened 2017-04-05
Added storing active tab of form in cookies

Piterden:


Piterden:
How do you think, @RyanThompson do we need to reset cookies sometimes?

RyanThompson:
Ya - and it'd be nice to check in Twig so we didn't have to wait for JS to set the tab on page load

RyanThompson:
The cookie should be overridden by subsequent forms on other pages so only one form is kept and preferable only sticks if staying on the form (saving only - not save and exit)

Piterden:
Ya! It's a great idea about to check in Twig!

It store cookies depending on route now. For each form separately.
So we don't need to override it in subsequent form.

Piterden:
But maybe we set limit on count of cookies or something like this?

RyanThompson:
I think a naming convention similar to `\Anomaly\Streams\Platform\Ui\Table\Command\SaveTableState::handle` will work well.

Piterden:
Could I get it fast in js?

Piterden:
Sorry, I don't understand about convention.
Now, it is like
image
Please explain more detailed what do you mean.

PR #379, opened 2017-04-12
1.2

mattcdavis1:
This update allows for conditional twig caching.  This is useful for dev environment when you typically don't want to cache content - even if it is in cache tags.

### DIFF
--- a/src/View/Cache/CacheAdapter.php
+++ b/src/View/Cache/CacheAdapter.php
@@ -39,11 +39,7 @@ class CacheAdapter implements CacheProviderInterface
      */
     public function fetch($key)
     {
-        if (env('ENABLE_TEMPLATE_CACHING')) {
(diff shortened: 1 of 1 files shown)


RyanThompson:
Going to change this to TWIG_CACHE and default to true

mattcdavis1:
great - thanks!

mattcdavis1:
@RyanThompson - just pulled in this update.  Looks like you missed a find / replace [here](https://github.com/datacodesolutions/streams-platform/blob/83e066ce0e2adb521dbc83670a4fbf3447f414ac/src/View/Cache/CacheAdapter.php#L54)

RyanThompson:
Thanks Matt - got it!

mattcdavis1:
one last thing - looks like on that second twig cache it's not defaulting to true (line 55)

PR #381, opened 2017-04-18
Copy field slug button

Piterden:
image


Piterden:
We could copy something else with shift button pressed.

Piterden:
Have bug with nested forms now

RyanThompson:
I think this is best left out of the UI. There are the fields tables generally or assignments but.. idk I feel like while there is a need to know what's available - the people who need that should not be mucking up the UI for the end user. Docs generally have attributes. 

I'll have to think on a good way to expose this to the appropriate audience.

Piterden:
It is awesome! )

PR #386, opened 2017-05-06
1.2

mattcdavis1:
Don't escape special characters for field attribute values when compiling stream entry model.


mattcdavis1:
hmm - still escaping somewhere.  not ready yet.

mattcdavis1:
oh nm - should be fine.  forgot i had some twig stuff in there.  This should be good to go.

RyanThompson:
Thanks Matt!

PR #388, opened 2017-05-09
Event priority overwritten.

RichardMBS:
Value of $listener was being overwritten before being set to $priority.


RyanThompson:
Good catch, thank you!

RichardMBS:
No Problem, Thanks

PR #389, opened 2017-05-09
Allow developer to set their own form responses

ramcda:
Existing code overwrites all responses on ajax forms with generic


ramcda:
Doesn't work in v3.  Will rework.

PR #390, opened 2017-05-09
Override json response values with any custom response sent by form h…

ramcda:
…andler

AJAX Forms return one of 3 responses.  Null, Redirect, or a custom response
If we have a custom response (ie not null and not a redirect) override data with the custom response values.  This allows developers to pass back custom AJAX responses via their form handlers while still allowing defaults to flow through from the existing logic.


RyanThompson:
So what is `$original` in this case?

RyanThompson:
I don't see getData in the response family classes

PR #392, opened 2017-05-10
null check on migrator

### LINKED ISSUE #391: fatal error when running php artisan migrate:status
i get the following error on running artisan migrate:status
```
  [Symfony\Component\Debug\Exception\FatalThrowableError]  
  Call to a member function getAddon() on null   
```
this traces back to
https://github.com/anomalylabs/streams-platform/blob/master/src/Database/Migration/MigrationRepository.php#L29

The problem is that it doesn't check if migrator is still null.

mjrider:
check if migrator is null, and handle it correctly
fixes #391 
note: the fix was testen in an local vagrant box, but copy/pasted into the github editor

### DIFF
--- a/src/Database/Migration/MigrationRepository.php
+++ b/src/Database/Migration/MigrationRepository.php
@@ -26,7 +26,7 @@ class MigrationRepository extends DatabaseMigrationRepository
      */
     public function getRan($namespace = null)
     {
-        if ($addon = $this->migrator->getAddon()) {
+        if ($this->migrator !== null && $addon = $this->migrator->getAddon()) {
             return $this->table()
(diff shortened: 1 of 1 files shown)


RyanThompson:
Where / how did you run into this?

mjrider:
in an pyrocms installation, with custom addon/module/extentions(not sure about the right term) provided by an externe webdevelopment company.

after running the php artisan install, checking the migration status failed

RyanThompson:
I am going to take a look at an alternative for this. I feel like it shouldn't get here if things haven't been fulfilled prior.

RyanThompson:
Thanks for the heads up though!

PR #395, opened 2017-05-12
Return void for all EntryObserver events

mattcdavis1:
Returning a non-void value was causing issues with venturecraft/revisionable and is unneeded for any Pyro functionality.


RyanThompson:
Thanks Matt!

PR #396, opened 2017-05-12
Hotfix trans choice

agustindidiego:
trans_choice and trans_exists Twig Functions are throwing exceptions because $this->translator is not instantiated.


RyanThompson:
Doh!

PR #401, opened 2017-05-21
Add functionality for overriding entry stubs

mattcdavis1:
Per our slack discussion this pull request adds functionality so that entry stubs and entry stub data are overridable via an `entry.generating` event.  There was a number of different potential ways to handle this.  For example, another option would be to create an event class and use the return response (if provided) to override $data or $templatePath.  The approach i used seemed a bit cleaner because there's no "if this then that" logic - the `$config->get({item})` value is used in all cases.  

Additionally, i added a few values to the entry data configuration for future use (since entry.stub is outside the scope of this repo).  These items are: 'use', 'extends', 'traits', 'properties' and 'methods';  These extra options will reduce the probability of needing to replace the whole stub as you can do just about anything with those string replacements available.

Sample event handler might look like this:

```php
<?php namespace Datacode\DatacodeModule\Listeners\Entry;

use Anomaly\Streams\Platform\Stream\Contract\StreamInterface;
use Anomaly\Streams\Platform\Support\Collection;

class Generating
{
    public function handle(StreamInterface $stream, Collection $config)
    {
        $data = $config->get('data');

        $data['title'] = 'My Special Title';

        $config->put('data', $data);

        $config->put('templatePath', resource_path('stubs/entry.stub'));
    }
}
```


RyanThompson:
Sorry this took so long! But did it a little different: https://github.com/anomalylabs/streams-platform/commit/39bf026e502e181dee7044f6e71b25b04ae066b3

mattcdavis1:
great - thanks!

PR #402, opened 2017-05-25
Remember locale using localStorage

benrogmans:
For ease of use and to avoid confusion when users edit a page in a specific language and click 'update' in between.

### DIFF
--- a/resources/js/form/translations.js
+++ b/resources/js/form/translations.js
@@ -24,5 +24,7 @@ $(function () {
         localStorage.setItem('formTranslations', locale);
     });
 
-    $('[data-toggle="lang"][lang="' + localStorage.getItem('formTranslations') + '"]').first().click();
+    $(document).ready(function(){
+        $('[data-toggle="lang"][lang="' + localStorage.getItem('formTranslations') + '"]').first().click();
+    });
 });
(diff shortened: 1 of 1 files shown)


RyanThompson:
Thanks a ton! Nice little feature ^_^

Piterden:
Are there nested event listeners, or it's just my imagination?

RyanThompson:
How do you mean?

Piterden:
I mean `$(function ());` statement is actually, an alias to document ready event. We don't need to write another listener inside its callback function.

Piterden:
https://github.com/benrogmans/streams-platform/pull/1

Piterden:
Proof: https://api.jquery.com/jQuery/#jQuery3

Piterden:
Also, not be much excepted to check for availability and efficiency of localStorage in global namespace.
Suppose, the most right way would be writing of sessionStorage fallback, including the same check.

RyanThompson:
Ahh I see - good thinking we can clean those all up in a separate commit.

Piterden:
Ben could merge my PR, maybe it would be simpler?

benrogmans:
Done and created a new PR for the base repo

Piterden:
Thanks, Ben)

PR #403, opened 2017-05-25
StreamModel config data corrupted by Builder and EntryStreamParser.

RichardMBS:
- Removes unnecessary slashes for double quotes which are corrupting serialized data.
- Checks if $data['config'] is already a string before re-serializing it.

See [StreamModel config data corrupted by Builder and EntryStreamParser #4373](https://github.com/pyrocms/pyrocms/issues/4373)


RyanThompson:
Thanks!

PR #405, opened 2017-06-01
add LT valitaion

huglester:


RyanThompson:
Thanks! Keep em coming!

PR #408, opened 2017-06-04
Update StreamRepository.php

Piterden:


RyanThompson:
Wrong branch :-/ I usually handle CS in batches, thanks!

PR #410, opened 2017-06-04
Patch 22

Piterden:


Piterden:
Streams commands now not override native Laravel, like it was earlier.

Piterden:
`getOption()` on input throws an exception. Need to use `hasOption()` instead.

PR #413, opened 2017-06-07
Prevent upper cased arguments

Piterden:


RyanThompson:
I'll slugify it. Thanks!

PR #414, opened 2017-06-14
Fixes parsing issues.

RichardMBS:
- Adds `setRules()` method to `FieldType` class.
- Fixes various issues parsing objects to PHP code.

### DIFF
--- a/src/Addon/FieldType/FieldType.php
+++ b/src/Addon/FieldType/FieldType.php
@@ -298,18 +298,6 @@ class FieldType extends Addon
     }
 
     /**
-     * Set the rules.
-     *
-     * @param array $rules
-     * @return $this
-     */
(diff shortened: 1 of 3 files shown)


RyanThompson:
Thanks for this!

We've got a `mergeRules` method on the FT already. I don't know that I want a `setRules` because any rules specified by the FT should not be able to be removed IMO. That will cause issues.

For parsing object - can you explain to me some of what you were experiencing that needed fixed? So I can replicate it over here? Just curious more than anything.

RyanThompson:
If anything let's investigate/consider the changes to FT in another PR. Try and keep them to 1 particular fix.

Thanks again! Ready to merge after that ^_^

RichardMBS:
I have just replied to your email with the details.

PR #416, opened 2017-06-14
Refactored

Piterden:


RyanThompson:
Why? There's a lot of changes in there..

Piterden:
There's no changes. Script is doing absolutely the same. Just not share vars to the global scope and some shorter.

PR #418, opened 2017-06-16
[Streams-Platform] Skip empty lines from .env files #4406

agustindidiego:
- Added FILE_SKIP_EMPTY_LINES flag when loading .env file


RyanThompson:
Thanks!

PR #423, opened 2017-07-05
Prevent of incorrect addon namespace launch

Piterden:


RyanThompson:
This is done basically right under where your code block is. If no addon - yelp about it.

Piterden:
Just there was one guy, who wrote something like `flatpickr-field_type.field_type.datetime`.
And I thought that it wouldn't be superfluous to have a bit of help there.

RyanThompson:
LOL so it did indeed find the addon?

PR #424, opened 2017-07-05
Fix AppendEntitySingletons

markdelf:
ucfirst was causing a bad path for the ServiceProvider - now inline with other commands with studly_case


RyanThompson:
Good catch! Thanks

PR #428, opened 2017-07-15
More exact matching, safe arrays, more informative Exception

Piterden:


Piterden:
What's wrong?

RyanThompson:
This class needs more thought in general - don't wanna move on it just yet.

PR #435, opened 2017-08-13
Let it be formatted out from the box, maybe?

Piterden:


huglester:
In general all this formatting is incorrect and it does not comply with PSR2 :(

I'd like pyrocms to fix this issue..
this is generally a bad practice, when you add one LONG field, and come up with commit like this with 20 lines:
image

I think for now it can be closed. I will push @RyanThompson on solving PSR2 standards all over the packages once and forever :)

Piterden:
The thing is that you shouldn't have any long lines in that place.
I just made the code style, to the style, which I saw inside other packages of the system.
Don't know about PSR-2, but for me this task is doing by `php-fmt` formatter.

@huglester @cyrildewit
You guys could take a look, how it was done in other parts of the system: 
`vendor/anomaly/streams-platform/src/StreamsServiceProvider.php`

huglester:
Yeah I know Denis. I just commented that this should be fixed once and everywhere:D

RyanThompson:
How is this not PSR2?

PR #436, opened 2017-08-13
Append prepend method ))

Piterden:


RyanThompson:
Merged manually cause of the extras in the PR - thanks!

Piterden:
Eh...

RyanThompson:
Try and not change comments within functional PRs - isolate changes better - then if you need - submit another PR for CS.

But the link for example should not be dropped below - I have an automated system that does all that so it's just extra blah from your CS sniffer likely

PR #439, opened 2017-08-23
Hold shift to select checkmarks between clicks for selecting rows

Piterden:
When there are no one or only one selected, it toggles selection or select another element.
When there are multiple items selected it toggles only one leaving to other its state.


Piterden:
Check the JS. There is no one DOM modification.

Piterden:
@RyanThompson could you show what behavior you don't like?

RyanThompson:
Meh

Piterden:
What `meh` means?

Piterden:
Hm. What's again?

PR #440, opened 2017-08-23
Add polyfill for L5.4 UrlGenerator formatScheme method

fryiee:
Adds a polyfill for the L5.4 `formatScheme` method which has been utilised in the latest commits to the Url Field Type. See issue here: https://github.com/pyrocms/pyrocms/issues/4537 and previous PR directly on FT here https://github.com/anomalylabs/url-field_type/pull/10

### DIFF
--- a/src/Routing/UrlGenerator.php
+++ b/src/Routing/UrlGenerator.php
@@ -170,4 +170,22 @@ class UrlGenerator extends \Illuminate\Routing\UrlGenerator
     {
         return (bool)$this->routes->getByName($name);
     }
+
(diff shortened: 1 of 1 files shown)


RyanThompson:
We don't wanna get into polyfilling I dont think :-/

fryiee:
@RyanThompson all good, forgot about forceSchema being the pre 5.4 method.

fryiee:
@RyanThompson your patch did not correct this issue, as forceSchema and formatScheme are not synonymous.

PR #441, opened 2017-08-23
Adjusted the IF statement to require the .version() and config to be true

calebsmithdev:
If the config and `.version()` doesn't exist, or is true, then version the image. Otherwise, don't version the image.


RyanThompson:
Thanks!

PR #449, opened 2017-09-01
Removed autofocus

Piterden:
It triggered scroll to top, which is hellish annoying.

https://www.w3schools.com/tags/att_input_autofocus.asp

### DIFF
--- a/resources/js/form/form.js
+++ b/resources/js/form/form.js
@@ -2,9 +2,6 @@ $(function () {
 
     var $form = $('form.form');
 
-    // Focus on the first input.
-    $form.find('input:visible').first().focus();
-
     // Highlight tabs containing errors.
     $form.find('.has-danger').each(function () {
 


RyanThompson:
Where's it doing this behavior? Seems a common sense that the top of the page has the first input in visibility. Is it focusing on something wrong?

Closing for now but let's hear more about it (house cleaning)

Piterden:
Hey, then calculate how much times you create and how much times you edit the form. When I edit, I don't need that focusing

Piterden:
I have long pages and have to scroll again, after I edit and check front of page

Piterden:
The best behaviour would be to save focused field to the 🍪 on save

RyanThompson:
And reset that when you leave the form? like back to the table and then back to the form another entry form and it’ll default to the first?

Piterden:
It shouldn't be at the first input when you are entering the form in `edit mode`. It is very rarely case, when you would need to edit a title of a page in `edit mode`.

So, I propose at the first, to disable autofocus in the edit form mode. It is enough quick task.
Then let's think about the saving focus state behavior.

RyanThompson:
How does it not make sense on edit? Or.. how does it not not make sense.. it’s such a subtle convenience Im afraid we’re discussing the details of such a tiny thing.

Can you send me a video of your issue perhaps?

Piterden:
What is not clear for you? I have many fields assigned to the page. When I am editing some wysiwyg content, placed in the wysiwyg field at the bottom part of the page, I need to control this process in the frontend. Frontend I have opened in the neighbor tab of the browser. So, when I saving for checking how it would looks like in the frontend, I don't want the native behavior of the browser would be overriden. I want to be on the same place of the page after reload. Because scrolling is unnecessary here. Why every time after reload I again at the top. I have edited title long time ago and I don't want to edit it again. I have the correct title! 
Do you really need a video for understand it?

Piterden:
Anyway after any save of the page I need to check it. In the most of cases, I need to correct something else there after checking.

Piterden:
UP ^

PR #451, opened 2017-09-01
Added missing use statement

Piterden:


RyanThompson:
Actually this class is garbage.. deleted it. I was gonna replace the package and this was a start that never shoulda made it up into the repo.

PR #455, opened 2017-09-01
Added typehinting

Piterden:


RyanThompson:
Ya I usually don't do this cause it can be slightly dangerous with decoration.. Will be closing these :-|

PR #458, opened 2017-09-01
Fixed typehinting, added comments

Piterden:


RyanThompson:
These comments are unnecessary because the `return` on the method is documented.

PR #462, opened 2017-09-01
Fixed logic

Piterden:


RyanThompson:
What's wrong with the "logic" here?

PR #463, opened 2017-09-01
Refactored

Piterden:


RyanThompson:
I like this one the way it is - a little more verbose never hurt some times :-)

PR #464, opened 2017-09-01
Fixed make cache table command

Piterden:


RyanThompson:
Ima put this in core - it should come out of the box already made

Piterden:
What you mean core? The core folder?

RyanThompson:
https://github.com/anomalylabs/streams-platform/blob/1.3/migrations/application/2015_03_15_171509_create_cache_table.php <https://github.com/anomalylabs/streams-platform/blob/1.3/migrations/application/2015_03_15_171509_create_cache_table.php>

I try and have all the laravel tables made readily available.

PR #468, opened 2017-09-05
Fixed unnecessary calling of the email value

Piterden:


Piterden:
Ya I understand about excess checking

RyanThompson:
Not the responsibility of this class!

Piterden:
Maybe you could explain a little bit more detailed?

RyanThompson:
SRP - By the time the target gets here we should know it’s resolvable. 

I think the fix should go in settings / preferences.

PR #472, opened 2017-09-08
|| and && should always placed at the end of the line

Piterden:


RyanThompson:
Says who?

Piterden:
Linter sad.
```php
if (
    myCondition1 &&
    myCondition2 ||
    myCondition3
)
```
But
```php
return myCondition
    ? exp1
    : exp2
```

RyanThompson:
Which one?

RyanThompson:
Hmm.. it's tough: https://stackoverflow.com/questions/23758916/psr-2-standard-for-long-if-conditions

Piterden:
It would be better if we would have maximum similarity between JS and PHP code styles

RyanThompson:
Expanded the line anyways - used to be longer

PR #478, opened 2017-09-16
update LT message.php

huglester:
html fixes


huglester:
will you backport to 1.2 ar should I have done this? sorry.

PR #480, opened 2017-09-19
Added unified getter.

Piterden:
To be DRY. This behavior is repeated few times

### DIFF
--- a/src/Support/Getter.php
+++ b/src/Support/Getter.php
@@ -41,18 +41,66 @@ class Getter
      */
     public function get()
     {
+        return ($object = $this->checkId())
+            ? $object
+            : ($object = $this->checkSlug())
+                ? $object
+                : ($object = $this->checkPresenter())
+                    ? $object
+                    : ($object = $this->checkObject())
+                        ? $object
+                        : null;
+    }
+
+    /**
+     * Checks the id
+     *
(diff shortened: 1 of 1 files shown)


Piterden:
For example, the `GetMenu` command would looks like
```php
<?php namespace Anomaly\NavigationModule\Menu\Command;

use Anomaly\NavigationModule\Menu\Contract\MenuRepositoryInterface;
use Anomaly\Streams\Platform\Support\Getter;

/**
 * Class GetMenu
 *
 * @link          http://pyrocms.com/
 * @author        PyroCMS, Inc. <support@pyrocms.com>
 * @author        Ryan Thompson <ryan@pyrocms.com>
 */
class GetMenu
{

    /**
     * The menu identifier.
     *
     * @var mixed
     */
    protected $identifier;

    /**
     * Create a new GetMenu instance.
     *
     * @param mixed $identifier
     */
    public function __construct($identifier)
    {
        $this->identifier = $identifier;
    }

    /**
     * Handle the command.
     *
     * @param  MenuRepositoryInterface $menus
     * @return MenuInterface|null
     */
    public function handle(MenuRepositoryInterface $menus)
    {
        return (new Getter($menus, $this->identifier))->get();
    }
}
```

PR #481, opened 2017-09-22
Module not properly uninstalled when in prod env.

josbarbosa:
Missing force flag to clean up migrations.


RyanThompson:
Nice catch - thanks!

PR #484, opened 2017-09-25
The depends on the own environment for every each developer

Piterden:


RyanThompson:
Yep and they can adjust as needed

PR #487, opened 2017-09-25
Fixed getTablePrefix

Piterden:


RyanThompson:
Not exactly a getter.

PR #498, opened 2017-09-29
Add --tree option to the stream generator

Piterden:


Piterden:
So, the command
```bash
$ artis make:stream comments defr.module.catalog --tree
```
now generates next:
```php
<?php namespace Defr\CatalogModule\Http\Controller\Admin;

use Defr\CatalogModule\Comment\Form\CommentFormBuilder;
use Defr\CatalogModule\Comment\Tree\CommentTreeBuilder;
use Anomaly\Streams\Platform\Http\Controller\AdminController;

class CommentsController extends AdminController
{

    /**
     * Display an index of existing entries.
     *
     * @param CommentTreeBuilder $table
     * @return \Symfony\Component\HttpFoundation\Response
     */
    public function index(CommentTreeBuilder $table)
    {
        return $table->render();
    }

    /**
     * Create a new entry.
     *
     * @param CommentFormBuilder $form
     * @return \Symfony\Component\HttpFoundation\Response
     */
    public function create(CommentFormBuilder $form)
    {
        return $form->render();
    }

    /**
     * Edit an existing entry.
     *
     * @param CommentFormBuilder $form
     * @param        $id
     * @return \Symfony\Component\HttpFoundation\Response
     */
    public function edit(CommentFormBuilder $form, $id)
    {
        return $form->render($id);
    }
}
```
and I suppose the error is caused of wrong params for tree builder
```php
<?php namespace Defr\CatalogModule\Comment\Tree;

use Anomaly\Streams\Platform\Ui\Tree\TreeBuilder;

class CommentTreeBuilder extends TreeBuilder
{

    /**
     * The tree segments.
     *
     * @var array|string
     */
    protected $segments = [];

    /**
     * The tree buttons.
     *
     * @var array|string
     */
    protected $buttons = [];

    /**
     * The tree options.
     *
     * @var array
     */
    protected $options = [];

    /**
     * The tree assets.
     *
     * @var array
     */
    protected $assets = [];

}
```

RyanThompson:
Nicely done! Thank you ^_^

RyanThompson:
PS I did rename `--nested` .. makes more sense to my ears I guess.

Piterden:
I think about the definition of it. I mean, **the main list** view should have the definition. Maybe **list view** or something similar.

`--nested` is really unclear for me, sorry...

Piterden:
I mean we have 2 variants of the list view showing method:

- table
- tree

And then there is nesting?

PR #500, opened 2017-10-03
Update Presenter.php

aliselcuk:
why is this checking the method on presenter but trying to call it on the object itself?

is this a typo? or is it intended so


Piterden:
Good catch! )

PR #501, opened 2017-10-03
needed when using twig templates in queue jobs

aliselcuk:
In queue jobs theme is not set, so this is needed to avoid an exception on $this->theme->* related calls


aliselcuk:
has ryan stopped merging PR's ?

PR #509, opened 2017-10-10
Fixed safe arrays

Piterden:


RyanThompson:
Hey man - I think these kinds of PRs should be avoided. I like the native array operations for the most part.. they tend to be faster / more readable. I usually only use those helpers when it condenses code or improves readability in some way. On top of that it's hard to tell if nuance functionality has changed at all without tests (which we're working on). But I am afraid I'll have to close similar PRs to this as they are not fixing anything and could cause edge case issues to crop up.

Piterden:
Had you tested the performance differences between em?

PR #510, opened 2017-10-11
Removed `null` initial value

Piterden:
I believe they anyway would be null at class init.


RyanThompson:
Null is intended. They only have no default value if expected by the `__construct()`

Piterden:
image

https://repl.it/OHaw/1

Piterden:
Reopen it @RyanThompson !!!

PR #511, opened 2017-10-19
Fixed an exception when getting rules

Piterden:
```php
    /**
     * Get the assignment's field's rules.
     *
     * @return array
     */
    public function getFieldRules()
    {
        $field = $this->getField();

        return $field->getRules();
    }
```
getRules is undefined on a field.


RyanThompson:
Needed added to interface - did this manually - thanks!

PR #516, opened 2017-11-17
Removed annoying warning

Piterden:
> Warning: Ambiguous class resolution, "JSMin" was found in both "/PATH/vendor/lmammino/jsmin4assetic/src/JSMin.php" and "/PATH/vendor/linkorb/jsmin-php/src/jsmin-1.1.1.php", the first will be used.


RyanThompson:
What does this leave us for JS min?

Piterden:
`/vendor/linkorb/jsmin-php/src/jsmin-1.1.1.php`

RyanThompson:
HTMLCompress package brings that in - note to future-self

PR #517, opened 2017-11-18
1.2

RyanThompson:


Piterden:
Those global constants is the bad practice in JS!

PR #518, opened 2017-11-21
What were you doing here?

Piterden:


RyanThompson:
Sanitizing while leaving PRE/CODE blocks for display in things like docs n such.

Piterden:
What if there would be few `code` and `pre` tags? What if they would be nested?

RyanThompson:
Nice!

PR #519, opened 2017-11-23
Add Dutch translation

wearekonzept:


dwainscheeren:
Dutch translations are good! :)

PR #520, opened 2017-12-01
Fixed wrong path generation

Piterden:
It was `storage::.support.parsed.e692baaa8c7aec04ae0e108de6f03498` and `{{PATH}}/storage/streams/default//support/parsed/e692baaa8c7aec04ae0e108de6f03498.twig`

Now good


RyanThompson:
Good catch!

PR #522, opened 2017-12-29
Add more docblock info and default to route function

psaunders:
This would have saved me some time. Helps new users see what the route() function is supposed to do.


RyanThompson:
Thanks! I'll improve the doc but skip the default value ^_^

PR #525, opened 2018-01-15
Resolve pyrocms/pyrocms#4723 - Prevent namespaces with dashes

ddnetters:
See pyrocms/pyrocms#4723


Piterden:
Why don't replace dashes with lowdash?

PR #530, opened 2018-01-26
Update tabs.twig

fryiee:
This updates my change to ensure it covers both methods of slug definition (array key and the 'slug' parameter).


resources/views/form/partials/tabs.twig
```diff
@@ -2,7 +2,7 @@
     {% spaceless %}
         {% for slug, tab in tabs %}
             <li class="nav-item {{ loop.first ? ' active' }}">
-                <a href="#tab-{{ slug }}" data-toggle="tab" class="nav-link">
+                <a href="#tab-{{ tab.slug ? tab.slug : slug }}" data-toggle="tab" class="nav-link">
```
Piterden: `<a href="#tab-{{ tab.slug|default(slug) }}" data-toggle="tab" class="nav-link">`

resources/views/form/partials/tabs.twig
```diff
@@ -13,7 +13,7 @@
 
 <div class="card-block">
     {% for slug, tab in section.tabs %}
-        <div class="tab-pane {{ loop.first ? ' active' }}" id="tab-{{ slug }}">
+        <div class="tab-pane {{ loop.first ? ' active' }}" id="tab-{{ tab.slug ? tab.slug : slug }}">
```
Piterden: `<div class="tab-pane{{ loop.first ? ' active' }}" id="tab-{{ tab.slug|default(slug) }}">`

Piterden (changes requested):

resources/views/form/partials/tabs.twig
fryiee: Both acceptable, just depends how much Twig functionality we want to use. Not sure what would actually be quicker to render, @RyanThompson might know.

Piterden: `{{ tab.slug ?: slug }}` should work too...

RyanThompson:
Attll do! Thanks man!

PR #535, opened 2018-02-03
Add examples for frequencies with parameters

okaufmann:


docs/en/02.the-basics/07.service-providers.md
```diff
@@ -154,6 +154,12 @@ The `$schedules` property let's you easily define scheduled tasks.
         'daily' => [
     		\Anomaly\LogsModule\Console\ArchiveLogs::class,
     	],
+        'dailyAt|11:00' => [ // dailyAt('13:00')
+            \Anomaly\LogsModule\Console\ArchiveLogs::class,
+        ],
+        'twiceDaily|1, 13' => [ // twiceDaily(1, 13)
```
Piterden: What means `(1, 13)`? It is not clear.

okaufmann: It is run twice a day. Once at 01:00 and once at 13:00.
https://laravel.com/docs/5.5/scheduling#schedule-frequency-options

PR #536, opened 2018-02-05
Resolve out of the container custom TableBuilder repository

vargvinter:
Resolve out of the container for real.

Otherwise check in https://github.com/anomalylabs/streams-platform/blob/40fb6fc7f1dcc9a6fafb983e95ab99034a0babc7/src/Ui/Table/Command/GetTableEntries.php#L65 is always false.

### DIFF
--- a/src/Ui/Table/Command/GetTableEntries.php
+++ b/src/Ui/Table/Command/GetTableEntries.php
@@ -56,7 +56,7 @@ class GetTableEntries
         /*
          * Resolve the model out of the container.
          */
-        $repository = $this->builder->getRepository();
+        $repository = app($this->builder->getRepository());
         
         /*
          * If the repository is an instance of


src/Ui/Table/Command/GetTableEntries.php
```diff
@@ -56,7 +56,7 @@ public function handle()
         /*
          * Resolve the model out of the container.
          */
-        $repository = $this->builder->getRepository();
+        $repository = app($this->builder->getRepository());
```
Piterden, line 59:
```php
if (is_string(($repository = $this->builder->getRepository()))) {
    $repository = app($repository);
}
```

vargvinter: Yes, I know. I was too fast and didn't check.

PR #537, opened 2018-02-05
1.3.1

vargvinter:
Resolve out of the container for real.

Otherwise check in
https://github.com/anomalylabs/streams-platform/blob/40fb6fc7f1dcc9a6fafb983e95ab99034a0babc7/src/Ui/Table/Command/GetTableEntries.php#L65
is always false.

### DIFF
--- a/src/Ui/Table/Command/GetTableEntries.php
+++ b/src/Ui/Table/Command/GetTableEntries.php
@@ -56,7 +56,7 @@ class GetTableEntries
         /*
          * Resolve the model out of the container.
          */
-        $repository = $this->builder->getRepository();
+        $repository = app($this->builder->getRepository());
         
         /*
          * If the repository is an instance of
--- a/src/Ui/Table/Command/GetTableEntries.php
+++ b/src/Ui/Table/Command/GetTableEntries.php
@@ -56,9 +56,12 @@ class GetTableEntries
         /*
          * Resolve the model out of the container.
          */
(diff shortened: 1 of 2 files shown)


src/Ui/Table/Command/GetTableEntries.php
```diff
@@ -56,8 +56,9 @@ public function handle()
         /*
          * Resolve the model out of the container.
          */
-        $repository = $this->builder->getRepository();
-        
+        if (is_string(($repository = $this->builder->getRepository()))) {
```
Piterden: Have you checked it?

Piterden:
```php
$repository = $this->builder->getRepository();

if (is_string($repository) && class_exists($repository)) {
    $repository = app($repository);
}
```

More readable

Piterden (changes requested):

src/Ui/Table/Command/GetTableEntries.php
vargvinter: Yup, works.

If `$repository` is not provided in `TableBuilder` then `EntryTableRepository` is used. Otherwise `$repository` is resolved out of the container correctly and can be used (if implements `TableRepositoryInterface` -> docs should be updated as well).

RyanThompson: Would you mind updating docs real quick?

PR #538, opened 2018-02-12
Only show the table footer when there's something to show

squatto:
First off, `{% if table.data.pagination.links %}` evaluates to true even when there aren't links, so the footer was showing even if there wasn't any pagination. This resulted in the per-page dropdown showing even when you only had one page of results. I fixed this by adding the `length` filter.

I also surrounded the entire output in a conditional to check if there's anything to actually show in the footer. If not...no footer.

### DIFF
--- a/resources/views/table/partials/footer.twig
+++ b/resources/views/table/partials/footer.twig
@@ -1,51 +1,53 @@
-<tfoot>
-<tr>
-    <th colspan="50%" style="padding: 10px;">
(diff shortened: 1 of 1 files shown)


RyanThompson:
Thanks! I ended up mirroring this myself from your commit cause it looks like there are formatting issues? With spaces vs tabs maybe Idk. Thanks though! Nice polishing fix here.

PR #539, opened 2018-02-14
Allows custom messages per field/rule

edster:
Before was setting generic message for all fields using a given rule.

Change allows individual messages per fields rule.


edster:
https://laravel.com/docs/5.6/validation#custom-error-messages

RyanThompson:
Hell ya

PR #541, opened 2018-02-16
Improved de translation

mmodler:


mmodler:
I should not commit late evenings :)

PR #542, opened 2018-02-26
Feature addon loader

fryiee:
This PR adds support for enabled addons (such as modules, extensions) to override language files. Open to discussion on whether we should also check for current active themes, as these are also often used for language file overrides.


RyanThompson:
Nicely done! I appreciate the test too man, thanks.

PR #544, opened 2018-02-27
Added restrict on relation

Eighke:
I also added a bit of docs

### DIFF
--- a/src/Model/Command/RestrictDelete.php
+++ b/src/Model/Command/RestrictDelete.php
@@ -1,0 +1,68 @@
+<?php namespace Anomaly\Streams\Platform\Model\Command;
+
+use Anomaly\Streams\Platform\Message\MessageBag;
+use Anomaly\Streams\Platform\Model\EloquentModel;
+use Illuminate\Database\Eloquent\Relations\Relation;
+use Illuminate\Translation\Translator;
+
+/**
+ * Class RestrictDelete
+ *
+ * @link   http://pyrocms.com/
+ * @author PyroCMS, Inc. <support@pyrocms.com>
+ * @author Ryan Thompson <ryan@pyrocms.com>
+ */
+class RestrictDelete
+{
+    /**
+     * The eloquent model.
+     *
+     * @var EloquentModel
+     */
+    protected $model;
+
+    /**
+     * Create a new RestrictDelete instance.
+     *
+     * @param EloquentModel $model
+     */
+    public function __construct(EloquentModel $model)
+    {
+        $this->model = $model;
+    }
+
+    /**
+     * Handle the command.
+     *
+     * @param MessageBag $messages
+     * @param Translator $translator
+     *
+     * @return bool
+     */
+    public function handle(MessageBag $messages, Translator $translator)
+    {
+        foreach ($this->model->getRestrict() as $relation) {
+
+            /* @var Relation $relation */
+            $relation = $this->model->{$relation}();
+
+            if (method_exists($relation, 'withTrashed')) {
+                $relation = $relation->withTrashed();
+            }
+
+            if ($count = $relation->count()) {
+
+                $messages->warning(
+                    $translator->trans('streams::message.delete_restrict'),
+                    [
+                        'name' => $this->model->getTitle()
+                    ]
+                );
+
+                return false;
+            };
+        }
+
+        return true;
+    }
+}
--- a/src/Model/Command/RestrictDelete.php
+++ b/src/Model/Command/RestrictDelete.php
@@ -41,7 +41,7 @@ class RestrictDelete
      */
     public function handle(MessageBag $messages, Translator $translator)
     {
-        foreach ($this->model->getRestrict() as $relation) {
+        foreach ($this->model->getRestricts() as $relation) {
 
             /* @var Relation $relation */
             $relation = $this->model->{$relation}();
@@ -50,19 +50,17 @@ class RestrictDelete
                 $relation = $relation->withTrashed();
             }
 
-            if ($count = $relation->count()) {
+            if ($relation->count()) {
 
                 $messages->warning(
                     $translator->trans('streams::message.delete_restrict'),
-                    [
-                        'name' => $this->model->getTitle()
-                    ]
+                    ['name' => $this->model->getTitle()]
                 );
 
-                return false;
+                return true;
             };
         }
 
-        return true;
+        return false;
     }
 }
--- a/src/Model/EloquentModel.php
+++ b/src/Model/EloquentModel.php
@@ -93,7 +93,7 @@ class EloquentModel extends Model implements Arrayable, PresentableInterface
      *
      * @var array
      */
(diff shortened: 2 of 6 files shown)


src/Model/Command/RestrictDelete.php
```diff
@@ -0,0 +1,68 @@
...
+     * @param Translator $translator
+     *
+     * @return bool
+     */
+    public function handle(MessageBag $messages, Translator $translator)
+    {
+        foreach ($this->model->getRestrict() as $relation) {
+
+            /* @var Relation $relation */
+            $relation = $this->model->{$relation}();
```
Piterden: I think it would be better to use the `call_user_func*` function to do it...

src/Model/Command/RestrictDelete.php
```diff
@@ -0,0 +1,68 @@
...
+
+            if (method_exists($relation, 'withTrashed')) {
+                $relation = $relation->withTrashed();
```
Piterden, line 50: Is there the same `/* @var Relation $relation */` ???

src/Model/Command/RestrictDelete.php
```diff
@@ -0,0 +1,68 @@
...
+            }
+
+            if ($count = $relation->count()) {
```
Piterden: Why do you need a variable here?

src/Model/Command/RestrictDelete.php
```diff
@@ -0,0 +1,68 @@
...
+
+                $messages->warning(
+                    $translator->trans('streams::message.delete_restrict'),
+                    [
+                        'name' => $this->model->getTitle()
```
Piterden: Trailing comma! But I would write like `['name' => $this->model->getTitle()]` instead..

src/Model/EloquentModel.php
```diff
@@ -79,6 +79,13 @@ class EloquentModel extends Model implements Arrayable, PresentableInterface
      */
     protected $cascades = [];
 
+    /**
+     * The restricting delete-able relations.
+     *
+     * @var array
+     */
+    protected $restrict = [];
```
Piterden: I suppose it should be `$restricted` or `$restricts`. But it is better to ask Ryan about it...

src/Model/EloquentObserver.php
```diff
@@ -87,6 +88,10 @@ public function updatedMultiple(EloquentModel $model)
      */
     public function deleting(EloquentModel $entry)
     {
+        if (!$this->dispatch(new RestrictDelete($entry))) {
```
Piterden: Maybe it is better to use comparing with true here and return inverted from the command?

src/Model/EloquentObserver.php
```diff
@@ -87,6 +88,10 @@ public function updatedMultiple(EloquentModel $model)
...
+            return false;
```
Piterden: I guess just return would be enough. It is not JS ))

Piterden (changes requested):

src/Model/EloquentObserver.php
Eighke: You need to `return false` to cancel the detele ;)

src/Model/EloquentObserver.php
Eighke: Sound good :)

src/Model/Command/RestrictDelete.php
Eighke: I was drunk, already fixed but didn't push yet

src/Model/EloquentObserver.php
Piterden: Hm... No! You don't need to return anything from an observer method.

src/Model/Command/RestrictDelete.php
Eighke: It is. I'm just not sure if it is usefull to store it back.

But I used `https://github.com/anomalylabs/streams-platform/blob/1.3/src/Model/Command/CascadeDelete.php`, and it is used like that. :)

src/Model/Command/RestrictDelete.php
Eighke: Well same as above, it is based on CascadeDelete, so I used the same way :)

src/Model/Command/RestrictDelete.php
Piterden: ctrl+f `withTrashed`

src/Model/EloquentModel.php
Eighke: I guess `$restricts` is better, we already use `$cascades` for casdades. @RyanThompson your call.

src/Model/EloquentObserver.php
Eighke:
```
        if ($this->fireModelEvent('saving') === false) {
            return false;
        }
```

It is :)

Piterden: It is not! ))) I bet 200$

Eighke: Just tested, my brand was deleted, poor little brand. I'll send you my Paypal :D

Piterden: Man, try
```php
if ($this->fireModelEvent('saving') === false) {
    return;
}
```
And also
```php
if ($this->fireModelEvent('saving') === false) {
    return true;
}
```

Eighke:
```
      if ($this->fireModelEvent('saving') === false) {
            return false;
        }
```

Forgot to told you: It is part of the EloquentModel, it doesn't come from my code. It is what cancel the saving() and it is checking for `=== false`.

Eighke: So if the Observer doesn't return false, it will never return anything and the delete will continue. ;)

Piterden: > and it is checking for === false

Show me this checking in the sources.

> So if the Observer doesn't return false, it will never return anything and the delete will continue. ;)

The same would be if we return true or void!

Eighke: Lazy boy! :D

Eighke: Lazy boy! :D
https://github.com/anomalylabs/streams-platform/blob/8fc6c0ff9df75d2505e9038468edd65217503985/src/Model/EloquentModel.php#L427

Piterden: It only works for `saving`. You have another method.

`deleting`

Eighke: Oh wrong one, I should stop drinking...

https://github.com/laravel/framework/blob/13f732ed617e41608e4ae021efc9d13e43375a26/src/Illuminate/Database/Eloquent/Model.php#L780

Eighke: (it is like that since Laravel 5.0, no reason it has changed since ^^)

Eighke:
Rebased and changed from the comments

RyanThompson:
Very nicely done! Thank you

PR #545, opened 2018-03-26
Automatically add a namespace prefix to the HTTP controller

lilianjin:


lilianjin:
All along, the Pyro routing configuration is written in a full name, which results in a long name. It is expected that the method of constructing the short circuit will improve the efficiency of reading and development.

src/Addon/AddonServiceProvider.php
```diff
@@ -1,5 +1,6 @@
 <?php namespace Anomaly\Streams\Platform\Addon;
 
+use ReflectionClass;
```
Piterden, line 3: Laravel has the representation of this stuff, as I remember...

Try `app(get_class($this))` or `app()->make(get_class($this))`.

BTW, why not `static::class`? (not tested it!)

src/Addon/AddonServiceProvider.php
```diff
@@ -240,6 +241,21 @@ public function getListeners()
...
+	 * In addition, it is set as the URL generator's root namespace.
+	 *
+	 * @param  string $name
+	 * @return string
+	 */
+	public function getHttpNamepsace($name = '')
+	{
+		$name = $name !== '' ?  trim($name, '\\') : $name;
+		return (new ReflectionClass(get_class($this)))
+			->getNamespaceName() . '\\Http\\Controller\\' . $name;
```
Piterden, line 257: Read more sources, my friend )))

RyanThompson:
Are you saying this will make a marked improvement in performance? If so can you share your testing results? Otherwise I am keen to keep it the way it is.

RyanThompson:
I think I am going to close this for now. The reflection has always bothered me. It tends to be expensive and we do plenty of it already in the IoC container. Let's keep the concept in mind though. 4.0 will automate routing so maybe this will be less needed anyways!

PR #547, opened 2018-03-27
Update 06.service-container.md

madan2056:
Need to update laravel docs link, since document is updated in 5.5 compared to 5.3.


Piterden (approved):
:+1:

PR #551, opened 2018-04-03
Update Value.php

Eighke:
* The regex is the same, and if `$value` match `$this->template->render()` will return a `View`. So the presenter will never be triggered.
* Anyway, the presenter is rendered in `$this->template->render()`

### DIFF
--- a/src/Support/Value.php
+++ b/src/Support/Value.php
@@ -141,15 +141,6 @@ class Value
             $value = $this->template->render("{{ {$value}|raw }}", $payload);
         }
 
-        /*
-         * If the value matches a method in the presenter.
-         */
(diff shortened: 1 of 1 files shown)


RyanThompson:
Can you explain to me what this fixed? I am a little fuzzy. I don't think there are any issues in executing presenter methods in table columns / elsewhere. Can you give me a sue case of what this fixes?

Eighke:
It is not really a fix, just unsued code.

If you check a bit before, L140

https://github.com/anomalylabs/streams-platform/blob/d62786ce364e0a92a49e875a64161568ff2449f6/src/Support/Value.php#L140-L151

You already have exactly the same `if()`. So 2 cases:

- the if L140 is true, the `$value` will not be a string anymore but a View, and the if L147 will be false
- the if L140 is false, so the if L147 is mandatory false too

:)

PR #552, opened 2018-04-06
Update SetMetaInformation.php

fryiee:
Check that the model doesn't have timestamps disabled.


mmodler:
+1 Thanks for the fast PR!

fryiee:
This will not actually work as EloquentModel uses $timestamps = false by default, so more discussion is needed on the solution for this.

PR #553, opened 2018-04-06
Fix for the configuration keys being fucked up on WAMP/XAMP

edster:
Not sure if we need to line 79 with `str_replace('\\', DIRECTORY_SEPARATOR, $directory),` as well, I'm thinking yes but not sure?


RyanThompson:
Thanks man

PR #555, opened 2018-04-17
Fixed ajax actions

Eighke:
You forgot the ajax footer in this commit https://github.com/anomalylabs/streams-platform/commit/1c87f5fa29773e6e93a694cee4c06730d14dfd6a, so the actions are not working on ajax tables. :)


RyanThompson:
Thanks!

PR #565, opened 2018-04-30
Fix an exception thrown

Piterden:


RyanThompson:
Duplicate - thanks for lookin out though stranger!

RyanThompson:
Nvm Im stupid

RyanThompson:
It does have conflicts though homie

RyanThompson:
I don't think this should happen in the wild. If you have steps to reproduce please share otherwise Im closin for now!

PR #570, opened 2018-05-15
Refactors Kernel to use Request class instead of superglobals

adrianfalleiro:
This PR refactors the `Anomaly\Streams\Platform\Http\Kernel` class to use the Request class methods instead of the PHP superglobals.

I have moved the `$this->defineLocale()` and `$this->rewriteAdmin()` calls from the constructor and into a new `handle($request)` method, which runs them first, before passing the request onto the rest of the application.


RyanThompson:
I'll have to revisit this. But not merging as is. Will need to pick and choose as I review. Thanks!

PR #571, opened 2018-05-15
Replaces references to superglobal in FieldType with Request class methods

adrianfalleiro:
This PR refactors the `Anomaly\Streams\Platform\Addon\FieldType\FieldType` class to use the Request class methods instead of the PHP superglobals.

### DIFF
--- a/src/Addon/FieldType/FieldType.php
+++ b/src/Addon/FieldType/FieldType.php
@@ -7,6 +7,7 @@ use Anomaly\Streams\Platform\Support\Decorator;
 use Anomaly\Streams\Platform\Support\Presenter;
 use Illuminate\Database\Eloquent\Builder;
 use Illuminate\Database\Eloquent\Relations\Relation;
+use Illuminate\Http\Request;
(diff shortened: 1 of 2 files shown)


RyanThompson:
I can't change dependencies here :-/

RyanThompson:
For the time being I think this is fine. We'll have to audit the system in general at a higher level if we're going to remove all these. 

Out of curiosity - can you mind me again the reasoning behind this? Replacing variables with class usage?

adrianfalleiro:
@RyanThompson It's been a while since I looked at this but the motivation behind all of the PRs I submitted replacing superglobals with references to the Request class was to make PyroCMS more compatible with [PHP-PM](https://github.com/php-pm/php-pm). PHP-PM is a stateful PHP server (like  Node). 

In our rudimentary tests the page load time of our Pyro application dropped from 300-500ms to <100ms. 

The problem with superglobals is that in PHP-PM they do not get flushed on each request. So things like forms and query strings break :(

PR #573, opened 2018-05-18
Update MakeAddon.php

Piterden:
But better to use a Regex namings
https://regex101.com/r/cN06LA/1/

https://repl.it/@DenisEfremov1/KlutzyRareByte


RyanThompson:
Thanks stranger!

PR #575, opened 2018-05-20
Remove the repeated loading of jquery.min.js;

lnlion:
This causes all the jQuery plug-ins loaded in theme.js to fail.


RyanThompson:
Yep this is still in accelerant theme.

PR #579, opened 2018-06-08
Added config fields to an assignment form

Piterden:


RyanThompson:
This has been on my radar for so long but I am afraid it might cause issues. 

For example if the default value has a value on the field but is blank on the assignment.. The fallback is incorrect behavior now (assignment will use no default value). Same if the field value is changed - the assignment value will have SOMETHING cause of the form and therefore not fallback appropriately.

Piterden:
Hey, but values wouldn't be blank on assignment form. It renders that form already filled with values of a field form. If those values would be different, so they will be overridden. Or am I talking a bullshit.

Anyway this behavior is strongly necessary on my present production project. Guess what CMS we use)))

I care much about a handler field exclusion. Please don't close it for now. I will rethink a bit and make some changes there soon.

Piterden:
If the field value is changing then we should update all assignments in that field, and change there only the same values, which weren't be overridden by assignments. @RyanThompson 

Microsoft had bought the GitHub. Let's go to the GitLab...

PR #580, opened 2018-06-11
Optimized export

Eighke:
I finally found why the export was so slow using PyroCMS: The "automatic" eager loading from Relationship and Multiple fields.

The problem:

```
Entries: 17K+
Relations: 6 (with translations)

Result:

Fatal error: Maximum execution time of 300 seconds exceeded
```

The download doesn't even start  (with 300s timeout~).

----------------

Solution:

So I just disabled the eager loading (`$this->repository->getModel()->newQueryWithoutRelationships()->get()`).

The result is: the download start directly, and the export of the 17K entries finished without any problem (1.5Mo)

And we should not need the relations for a basic CSV export.

### DIFF
--- a/src/Http/Controller/EntryController.php
+++ b/src/Http/Controller/EntryController.php
@@ -143,7 +143,7 @@ class EntryController extends AdminController
         $callback = function () {
             $output = fopen('php://output', 'w');
 
-            foreach ($this->repository->all() as $k => $entry) {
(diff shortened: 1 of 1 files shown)


RyanThompson:
Thanks for digging into this one! I adapted this but put it into a repository method for now: https://github.com/anomalylabs/streams-platform/commit/63268776eb2a96c3f0f21291145c938a8f7d0723

Thanks!

Eighke:
Even better! Thanks :)

PR #582, opened 2018-06-13
Removed nobr tags

Piterden:
I was had to remove it cause of a mobile issue with buttons. It's better to use CSS for that behavior.

### DIFF
--- a/resources/views/table/ajax/header.twig
+++ b/resources/views/table/ajax/header.twig
@@ -15,17 +15,15 @@
         <th>
             {% if header.sortable %}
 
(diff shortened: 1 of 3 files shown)


dwainscheeren:
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/nobr

Piterden:
What you wanted to say?

dwainscheeren:
It needs to be removed indeed because it's not supported

RyanThompson:
Thanks @Piterden !

PR #584, opened 2018-06-18
remove unnecessary jquery.min.js

lnlion:
jquery has already loaded in theme.js


Piterden (commented):
Haha, and how would the sortable plugin work without jQuery? Or it had been imported in some other place? Cause if so, i suppose fokken assetic would exclude repeats, lefting only code you need. But I am not fully sure, of course...

Why did you comment it? I would remove it fully on your place...

lnlion:
see this https://github.com/anomalylabs/streams-platform/commit/3488a5564a4dcc98659b92e623a6ab049a4852bd

lnlion:
and in accelerant-theme; you've already loaded jquery.min.js

RyanThompson:
I removed this a bit ago and it's doing fine (forgot all about it). Thanks!

PR #585, opened 2018-06-18
Remove the repeated loading of jquery.min.js

lnlion:
jquery.min.js has already been loaded in accelerant theme, see https://github.com/anomalylabs/streams-platform/pull/575;


RyanThompson:
Forgot I did this a while ago after confirming!

PR #586, opened 2018-06-19
Clarified using dynamic selects in filters

rickmills:
Hopefully theres a much better way than what I'm suggesting here, and if so would appreciate it being documented as examples posted around don't work (e.g https://pyrocms.com/forum/channels/everything/how-to-render-the-filter-options-from-database-table-in-table-builder-filter-options).

Whilst what I'm suggesting is certainly not ideal, for now this seems like the only option available as far as I can tell.

### DIFF
--- a/docs/en/04.ui/07.tables.md
+++ b/docs/en/04.ui/07.tables.md
@@ -238,6 +238,33 @@ The query class must accept the query `Builder`, `Filter` instance, and is calle
 
 <div class="alert alert-primary">**Pro Tip:** Even automated stream filters can be completely overridden. Try providing your own custom query!</div>
 
+##### Dynamic Filter Options
+
+If you're creating a filter that uses a select box, your drop-down options can be either populated by a simple array, or dynamically. However to populate the fields dynamically, you'll need to create a constructor to perform your query as it's not currently possible to use a handler here.
+
+Your filter array would look something like this:
+
+    protected $filters = [
+        'status' => [
+            'placeholder' => 'Order Item Status',
+            'filter' => 'select',
+            'type' => 'select',
+            'query' => StatusFilterQuery::class,
+            'options' => [] // Note that this is intentionally empty.
+        ]
+    ];
+
+and your constructor would look like this:
+
+    public function __construct(Table $table, ItemStatusRepositoryInterface $status)
+    {
+        parent::__construct($table);
+
+        $this->filters['status']['options'] = $status->all()->pluck('name', 'id')->all();
+    }
+    
+As we've used the repository interface, any translations on your `name` field would be automatically handled.
+
 ##### The Filter Definition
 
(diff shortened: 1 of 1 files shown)


docs/en/04.ui/07.tables.md
```diff
@@ -238,6 +238,33 @@ The query class must accept the query `Builder`, `Filter` instance, and is calle
...
+        ]
+    ];
+
+and your constructor would look like this:
+
+    public function __construct(Table $table, ItemStatusRepositoryInterface $status)
+    {
+        parent::__construct($table);
+
+        $this->filters['status']['options'] = $status->all()->pluck('name', 'id')->all();
```
Piterden, line 263: This is bullshit! You don't need to do it. Creating such of a constructor is the bad practice. You need to create a new class for your filter, which should extends one of filter classes of platform.

Also `$this->filters['status']['options'] =` is the very-very bad practice too!!!

`Anomaly\Streams\Platform\Ui\Table\Component\Filter\Type\SelectFilter::setOptions()`

rickmills: Firstly whats with the attitude? No need to be an ass.

If you'd read my message you'd have seen that having researched, all official solutions provided do not work, nor are there any documented examples. I'm well aware that what I provided was an awful option, but given nothing else is working, and people have tried to get this working for a while, anything was better than nothing at this point.

Nice way to treat people attempting to help, if thats the general attitude around here I can see why theres so few people willing to help out.

You can run `setOptions` from the array. This was purely for situations where you are 100% using the table builder, and want an option field populated without having to mess around creating controllers (Which would completely go against the purpose of how its usage is explained in the streams documentation and videos).

If you have a better solution, please do provide it I genuinely want to improve the docs here with whatever method works best directly in the table builder. The ideal solution would be if you could push the options handling to a handler class, but that doesn't work at all, as has been explained in a couple of forum threads already.

You may wish to have a read through this... https://pyrocms.com/code-of-conduct

RyanThompson: @rickmills first let me apologize on behalf of the project. This behavior is 100% not acceptable behavior and I appreciate you helping out! I've taken appropriate measures to ensure this kind of attitude is no longer an issue from him as well.


That said - dynamic options in a select filter can be handled in a couple ways:

- You can use a `Filters` handler and set them there with a closure when you set the filters.

- You can/should be able to use a handler for options in the filter:

```
    'status' => [
            'filter'  => 'select',
            'options' => 'Example\TestClass@handle',
        ],
```

I REALLY appreciate you digging into the docs. It's a sore spot for us and the help is huge! Take a look at these approaches and let me know what you think about them. I spot tested the handler and it's working so you might have had something off. Check it out though!

I also apologize for not getting to this sooner, for multiple reasons. I was out of town for a week and have been catching up and reorganizing efforts.

PR #588, opened 2018-06-19
Added info for setting the default table sort order

rickmills:
Quick little update to the options table to add in the missing default sort order.


RyanThompson:
This is conflicting cause of the other PR I merged - so I added it manually. Thanks!

PR #599, opened 2018-10-01
Allow to for the default locale route

Eighke:
Currently the URL with the default locale or without are the same (e.g. `/services` and `/en/services`) not the best for SEO. And most of the customers don't want this kind of behaviour.

So we can just add a settings to allow to change this behaviour.

* None (default), nothing change, both URL are avalaible
* 1, the URL without locale will redirect to the URL with
* 2, the URL with locale will redirect to the URL without


Eighke:
Doesn't work correctly, it broke admin route and other customer routes

RyanThompson:
I've got an idea about this I'll try out

PR #602, opened 2018-10-22
Allow listener during install

Eighke:
Enable the listener during `install`, allowing module to interact together

### DIFF
--- a/src/Installer/Console/Command/LoadModuleListeners.php
+++ b/src/Installer/Console/Command/LoadModuleListeners.php
@@ -1,0 +1,56 @@
+<?php namespace Anomaly\Streams\Platform\Installer\Console\Command;
+
+use Anomaly\Streams\Platform\Addon\Module\Module;
+use Anomaly\Streams\Platform\Addon\Module\ModuleCollection;
+use Illuminate\Contracts\Events\Dispatcher;
+
+/**
+ * Class LoadModuleInstallers
+ *
+ * @link   http://pyrocms.com/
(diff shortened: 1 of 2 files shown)


RyanThompson:
I can't say I am too stoked about this kind of coupling. There are better ways to do this IMO.

Eighke:
I don't see anyother way currently. I tried out many thing and it is the best one I could find (currently the best was this one but using a different `Provider` only for that, something like `getInstallerProvider()`).

Using the `Module::onInstalling` is possible but complicated. And you will need to use the `ServiceProvicer::$listerners` anyway.

-------------------------

Two module A and B

* Module A can interact on module B if this one is installed
* Module B is not aware of Module A

**Installing**

* Case 1: Installing using `php artisan install`
  * **Possible**: `A::onInstalled` to check if B exists and add `B::on('installed')`
  * **Not**: `AServiceProvider::$listerner` is not trigger
* Case 2: Installing B using `php artisan addon:install B`
  * **Possible**: Using `AServiceProvider::$listerner` to detect the installation of `B`
  * **Not**: `A::onInstalled` is not trigger

So we will need to duplicated the code in both `A::onInstalled` and `AServiceProvider::$listerner` to manage the 2 cases.

Also the `A::onInstalled` has 2 cases to manage depending on the name of A and B:

* If A is installed before B, we need to use a B::on('installed')
* If A is installed after B, we need to interect directly

-----------------

Well if I have missed something I'll be happy to find what. :)

PR #603, opened 2018-10-30
Asset library fails to resolve file system paths correctly

ipimpat:
Fixes an issue with resolving asset file system paths correctly.


ipimpat:
fixes https://github.com/pyrocms/pyrocms/issues/4863

RyanThompson:
As assetic is getting pulled out and the asset class get's cleaned up - I've fixed this in the 1.7 branch based on what you've provided me here. Thanks! Sorry I couldn't get this in sooner - I really wanted to let the asset class sit still until it could be given more attention.

PR #604, opened 2018-11-02
Don’t use the parent translation when value is “0”

weotch:
If the value of an attribute was 0, don’t default to the parent because “0” represents an explicit choice in boolean fields.  We ran into this when localizing switch-type boolean fields.


RyanThompson:
Excellent - thanks!

weotch:
Any chance you can release a 1.2.x version too?  That's where I forked from and what my project is on.  Thanks!

PR #605, opened 2018-11-20
add an event / callback after query

fryiee:
As title.


RyanThompson:
I dig it! I'll merge it after testing locally.

PR #613, opened 2018-12-14
Entries Fetched Listener

edster:
Allow you to interact with table entries that have been set from a model before creating the columns.


RyanThompson:
Found a GOOD alternative (override method)

PR #621, opened 2019-01-24
Update EloquentModel.php

lilianjin:
eplace get_called_class()  `get_called_class()` can be securely replaced with `static::class`. Also, this function will be marked as deprecated in PHP 7.4:  https://wiki.php.net/rfc/deprecations_php_7_4#get_called_clas


RyanThompson:
Thank you :-)

PR #631, opened 2019-02-25
Aliasing the route middleware

Eighke:
Route alias are not registered to the router. So using them in `public function map(Router $router)` is not working.

e.g.

```php
    ...
    protected $routeMiddleware = [
        'contents' => ContentMiddleware::class
    ];

    public function map(Router $router)
    {
            $router
                ->name('contents.show')
                ->middleware('contents')
                ->get('{contentSlug}', 'ContentsController@show');
    ...
```

Will fail

```
Class contents does not exist
```


RyanThompson:
Good looking out - thanks!

PR #632, opened 2019-02-25
fix php 7.3 support

schmidex:
compact() now issues an E_NOTICE level error if a given string refers to an unset variable. Formerly, such strings have been silently skipped. 

Don't know why group was even there?!


RyanThompson:
Oh man - so old!

RyanThompson:
Thanks!

PR #634, opened 2019-02-26
Remove legacy "fire" method and replace with "dispatch"

schmidex:
Hopefully in preperation of an laravel 5.8 upgrade where "fire" was removed


RyanThompson:
conflicts - but I am gonna replace with `event($event)`

PR #635, opened 2019-02-26
Cleanup BaseController

schmidex:
Not used (anymore)


RyanThompson:
Not sure what this does other than shortens use statements. I assume keep them consistent with everything else unless there is notable performance difference.

PR #637, opened 2019-03-01
try to improve PluginCriteria

schmidex:


RyanThompson:
Can you walk me through with maybe a bit of code as to what problems this solves?

src/Addon/Plugin/PluginCriteria.php
```diff
@@ -198,8 +213,8 @@ public function getCacheKey()
             $this->options,
             'cache.key',
             $this->getCachePrefix() . '.'
-            . md5(json_encode($this->collection)) . '.'
```
schmidex, line 201: `$this->collection` is nearly static. It hashes the collectionClass name (string). that nearly never changes. So no need to include this.

src/Addon/Plugin/PluginCriteria.php
```diff
@@ -218,13 +233,15 @@ public function getCacheKey()
     /**
      * Get the cache prefix.
      *
-     * @return null|string
+     * @return string
      */
     public function getCachePrefix()
     {
-        if ($model = $this->getModel()) {
```
schmidex, line 225: Why only include that when there is a model? If it is explicitly set, why not use it?

src/Addon/Plugin/PluginCriteria.php
```diff
@@ -247,7 +264,7 @@ public function setCachePrefix($cachePrefix)
      */
     public function getCacheTtl()
     {
-        return array_get($this->collection, 'ttl', 60 * 24 * 7);
```
schmidex, line 250: `this->collection` is a `string` not an `array`. It simply holde the name of the CollectionClass. So this can't return anything!

src/Addon/Plugin/PluginCriteria.php
```diff
@@ -258,7 +275,7 @@ public function getCacheTtl()
      */
     public function setCacheTtl($cacheTtl)
     {
-        $this->cacheTtl = $cacheTtl;
```
schmidex, line 261: There is no `$this->cacheTtl` in this class

src/Addon/Plugin/PluginCriteria.php
```diff
@@ -198,8 +213,8 @@ public function getCacheKey()
             $this->options,
             'cache.key',
             $this->getCachePrefix() . '.'
-            . md5(json_encode($this->collection)) . '.'
-            . md5(json_encode($this->options))
+            . md5(json_encode($this->options)) . '.'
+            . md5(json_encode($this->getCallbackStaticVariables()))
```
schmidex, line 217: `$this->getCallbackStaticVariables()` gets the variables that are passed through a `use` statement.
For example the `$menu` variable here: https://github.com/anomalylabs/navigation-module/blob/2.4/src/NavigationModulePlugin.php#L33
Or the `$method` variable here: https://github.com/anomalylabs/twitter-extension/blob/1.0/src/TwitterExtensionPlugin.php#L31

This is important, because these are the original twig plugin paramters.
I came across several cases where I got the same cache key for different requests because this was missing.

src/Addon/Plugin/PluginCriteria.php
```diff
@@ -21,13 +21,6 @@ class PluginCriteria
      */
     protected $model = null;
 
-    /**
-     * The cache prefix.
-     *
-     * @var null|string
-     */
-    protected $cachePrefix = null;
-
```
schmidex, line 30: I combined every cache relevant option in `$this->option['cache']` to keep it consistent. All other cache options (ttl, namespace, key) are already there.

src/Addon/Plugin/PluginCriteria.php
```diff
@@ -88,7 +81,17 @@ public function cache($parameters = null)
             ];
         }
 
-        $this->options['cache'] = $parameters;
+        array_set(
+            $this->options,
+            'cache',
+            array_merge(
```
schmidex, line 87: This has to be `array_merge` to NOT overwrite cache options that had been set by `$this->setCachePrefix()` or `$this->setCacheTtl()` before.

RyanThompson:
I see some good ideas here. I also have a key store method made up on the models now so we can easily bind this cache to models if present. Ima keep this around for reference though it won't be merged in verbatim. Thank you!

PR #640, opened 2019-03-06
Remove duplicate middleware

drmonkeyninja:
The cookie and session middleware are already being defined in the global middleware stack above. Therefore there's no need to redefine them for the web middleware group.


RyanThompson:
Thank you!

PR #641, opened 2019-03-06
Extend Laravel's VerifyCsrfToken

drmonkeyninja:
The addons `VerifyCsrfToken` class was mostly a copy of the Laravel class so it makes more sense to extend the class and add in the Pyro specifics to the extended class. This ensures any fixes that get added to Laravel will get passed through to Pyro in future.


RyanThompson:
Good one - thanks!

PR #643, opened 2019-03-06
Update EntryTranslationsModel.php

schmidex:
There is no `entry`. Fixes php 7.3


biggins:
@RyanThompson would it be possible to backport this to 1.5 as well?

PR #645, opened 2019-03-11
Allowed dedicated migrations from app

Eighke:
The current `migration` commands have a problem: when you use it with `--addon` it will ignore all `app` migrations.

If I create a generate module that I customize in the `app`, no migration will checked. It is really troublesome with `php artisan migrate:refresh --addon xx.module.yy` and working with the `app` migrations.

---------------

So I just added the default Laravel migration path + the addon namespace as second path (e.g. `database/migrations/xx.module.yy`).

Also added `realpath` from L5.6 since we are in 5.7.


RyanThompson:
Good one!

PR #651, opened 2019-03-21
Added card-body for BS4.0+

Eighke:
The final version of BS4 is using card-body instead of card-block.

I'm using both, so we can start moving on BS4.0+ and keeping the legacy for old themes.


Eighke:
(And we can remove the legacy class later, like on 1.7 or 1.8)

PR #652, opened 2019-03-21
Resolve rules when using handle

Eighke:
When creating  a `XxxFormRules`, `$rules` it is correctly set to `classname@handle`, but it's not resolved anywhere.


RyanThompson:
Nicely done!

PR #660, opened 2019-05-31
Allow icon view to be override

Eighke:


Eighke:
So we can override it on the theme. Actually we need to override class `Icon` itself

RyanThompson:
I don't think invoking the view system so often for a tiny fragment of HTML is a good idea. What did you need to override that we might be able to help with in another way?

Eighke:
Too complicated w/o the view, I'm closing. I just created a `svg` for svg Icon.

PR #667, opened 2019-07-15
Fixed config override for addons

Eighke:
The current config path is `resources/addons/namespace/addon-type` instead of `resources/addons/namespace/addon-type/config`.

And it will break if we override something else (like the views `resources/addons/namespace/addon-type/views`). Because it will try to load all the `.twig` as PHP config files. ^^;


RyanThompson:
Good one! BC or not this is correcting erroneous behavior in a Beta release I say we're golden - thank you!

PR #673, opened 2019-10-22
Manage MultipleForm prefix

Eighke:
Without MultipleForm prefix are ignored, and no custom message are added (else it need to redefine all fields in the MultipleForm :()


Eighke:
PS: Commit on 1.6 because I need it for my current project :)

PR #681, opened 2019-11-06
Update constants.twig

itinnovative:
APPLICATION_URL is not sanitized before output.  This creates a potentially serious XSS security threat.  Output of the URL is sent back inside a Javascript tag as-is which is dangerous.  Can easily be reproduced by turning of url encoding on your browser and hitting a URL such as "http://testsite.com/%22;alert('test');%3C!--"


RyanThompson:
Pulled this into 1.6 and merged it upward.  Thank you!

PR #683, opened 2019-12-16
Add initial JS and initializer view

RobinRadic:
What i currently have is:

- Streams Platform JS provides the
  - `Application`
    - Is a IoC Container
      - Extends the base class `Container` provided by the awesome library `InversifyJS`
      - I have added a few wrapper/alias functions to get a more 'Laravel'ish feeling like `alias()`, `bindIf()`, `singleton()`, `instance()`, etc
      - ServiceProviders are registered using `register()`
    - Contains the root `Vue` component (the component, not instance)
      - Can be extended by using `app.extendRoot({ ...vuestuff })`. This way Service Providers can add `data`, `methods` etc to the root component
      - Will be used to create the root `Vue` instance `app.root = new app.Root`
      - And will be mounted
    - Executes the lifecycle
      - Register all service providers
      - Boot all service providers
      - Initialize all Vue Plugins
      - Start the application (basically `app.root.$mount('#app')`)
  - `ServiceProvider`
    - Should be regarded as an abstract class, and should be extended
    - Can contain either/both `register` and `boot`. Both may also return a `Promise` if there's any `async` operations in it.
    - Has a easy short-hand `this.vuePlugin(plugin, options)` for adding vue plugins
    - Has the `Application` instance accesible `this.app`
  - `Dispatcher`
    - Primarily used by the `Application`, gives the ability to modify the lifecycle behaviour
  - Utility methods

The `init_platform.blade.php` thing is temporary until a better solution is agreed upon.
For now just add it to your `assets.blade.php` inside the flow theme or something. `@include('streams::init_platform')` at the eof.


RyanThompson:
Woo hoo! Thanks again! I am reviewing/learning :-)

PR #689, opened 2020-01-20
Hide dropdown buttons we don't have permission to use.

bubba169:
Currently the dropdown buttons are shown and link to a page that errors with a 403 status. To prevent confusion for end users, the dropdown button is now hidden. This is configured using the existing `permission` key in the buttons array.


RyanThompson:
Nice one! Thank you.

PR #692, opened 2020-03-19
Make this cookie support secure.

edster:


edster:
Going to change approach.

PR #697, opened 2020-07-06
Use `getFieldValue` for translated relationship attributes.

bubba169:
Skip trying to get the translated relationship using the laravel relationships as this ends in an error if the translated value is empty. It errors when serialising because it doesn't check for the `_id` field on the translation table, only the field slug. It also errors if a value is null as it tries to return the field from the parent using the original slug. 

Skipping this and going straight through `getFieldValue` will return the relationship id properly which is what currently happens if the translated values are present. This has the added benefit of fetching the default translated value (e.g. for `en`) if a translation for the current locale is not available.

Unless there's a better way to use translated relationships that I'm not aware of, this fix allows us to use the feature and work around the relationship issues by using a presenter to fetch the related model by its id.

### DIFF
--- a/src/Entry/EntryModel.php
+++ b/src/Entry/EntryModel.php
@@ -428,8 +428,12 @@ class EntryModel extends EloquentModel implements EntryInterface, PresentableInt
     public function getAttribute($key)
     {
 
-        // Check if it's a relationship first.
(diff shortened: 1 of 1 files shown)


RyanThompson:
Thank you!

PR #698, opened 2020-07-07
Use custom lexer to avoid minified CSS breaking twig comments in parse

fryiee:
This prevents an issue where minified CSS (usually produced by Mix) with any media queries (resulting in a line such as `@media (max-width:768px){#example{}}`) that was run through the `parse` filter would result in the "Unclosed comment" issue, which will break the page.


RyanThompson:
Looks good to me, thank you!

PR #701, opened 2020-08-12
Fixed MultiLanguage Problems

spektra2147:
getTranslationsTableName() removed to prevent data repeatation.The default addon language has been added for languages ​​that are not active in the system.

@fatihalp


fatihalp:
With adding and removing languages from admin panel causes difrrent erros which reported as github issues. 

Almost 1 year we're using that method with multi language sites and has no problem @RyanThompson

RyanThompson:
Thank you! I'll take it for a spin. Putting together a 3.8 right now.

PR #702, opened 2020-08-20
Fixes typo in how to get a form for manual rendering

finnito:
Error "Method Anomaly\Streams\Platform\Ui\Form\FormCriteria::__toString() must not throw an exception" is thrown if `.get()` is omitted


RyanThompson:
Thank you!

PR #726, opened 2021-01-11
Fix typo in namespace

giorgiopogliani:


fryiee:
@RyanThompson this is fine to merge also.

RyanThompson:
I suppose I should be watching these closer O_O

Thank you!

PR #727, opened 2021-01-12
Fixed stupid float behaviour

Eighke:
100 is not enough. 19.99 will become 19.98...

### DIFF
--- a/src/Support/Currency.php
+++ b/src/Support/Currency.php
@@ -52,7 +52,7 @@ class Currency
             $suffix = $this->symbol($currency);
         }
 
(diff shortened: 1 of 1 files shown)


RyanThompson:
Ugh, is this the "proper" way to handle this? I like the approach but got me thinking about the next version. Stupid behavior indeed! Thanks for the patch.

Eighke:
I saw on PHP docs that float are stupid. I see many are using `* pow($number, $decimal) / pow($number, $decimal)` ... but well I assumed we are only using 2 decimals max.

PR #728, opened 2021-01-16
Glue comes before array

edster:
glue must come before array in php 7.4


fryiee:
It's actually just deprecated in 7.4.0, but there's nothing wrong with this change. @RyanThompson good to merge?

RyanThompson:
Muchas gracias!

PR #729, opened 2021-02-09
temporary sass compile solution

spektra2147:
I noticed that scss files are not compiled. I used this solution temporarily for this.


RyanThompson:
Hello! We have purposefully removed PHP compilation of SASS/SCSS and other compiled languages like Coffee Scripts.

I would like to implement a filter registry where you can add this functionality to your project simply by registering the filter with the asset manager.

PR #745, opened 2022-12-27
fix installation error

spektra2147:
Contains solution for 'forge'@'localhost' error during installation.

fix1pyro


RyanThompson:
Pushed this in a little differently and cleaned up that class in general.

PR #750, opened 2023-11-15
upgrade Laravel 10

spektra2147:
--app parameters have been removed for installation issues. It is not available in the AddonInstall class.

SeedCommand added. Seeder not working.

Image class has been created.MountManager is not available.


spektra2147:
Laravel cache package was forked and the problems were solved.

PR #751, opened 2023-11-16
added composer package

spektra2147:
Class "Composer\Autoload\ClassMapGenerator" not found  

fixed

image


spektra2147:
The Composer\Autoload\ClassMapGenerator package will be deprecated and this commit includes the new edits.

